PR #32, opened 2025-07-01
Bug fix & Ux

738NGX:
- [x] BUG: File explorer with long name. Image svg is shrijking and ellipis text is overflowing.
- [x] UX: make more prominent that after 1st image, user can drag/upload more images

### DIFF
--- a/src/components/baseFileExplorer/baseFileExplorer.view.yaml
+++ b/src/components/baseFileExplorer/baseFileExplorer.view.yaml
@@ -77,12 +77,6 @@ styles:
     '-moz-user-select': 'none'
     '-ms-user-select': 'none'
     'user-select': 'none'
-  'rtgl-svg':
-    'flex-shrink': '0'
-  'rtgl-text[ellipsis="true"]':
-    'min-width': '0'
-    'flex': '1'
-    'margin-right': '24px'
 
 template:
   - rtgl-view w=f h=f: 
@@ -96,7 +90,7 @@ template:
                   - rtgl-view wh=16:
               - rtgl-view d=h g=md av=c w=f:
                 - rtgl-svg svg=${item.type} wh=16:
-                - rtgl-text s=sm ellipsis=true w=f: "${item.name}"
+                - 'rtgl-text s=sm ellipsis=true w=f flex=1 mr=xl style="min-width: 0"': "${item.name}"
       - $if targetDragIndex != -2 && targetDropPosition != 'inside':
           - 'rtgl-view pos=abs bgc=fg h=2 style="top: ${targetDragPosition}px; left: ${targetDragLeftOffset}px; width: calc(100% - ${targetDragLeftOffset}px); z-index: 1000;"':
     - rtgl-view#container w=f h=f:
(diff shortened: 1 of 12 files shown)


src/components/baseFileExplorer/baseFileExplorer.view.yaml
```diff
@@ -77,6 +77,12 @@ styles:
     '-moz-user-select': 'none'
     '-ms-user-select': 'none'
     'user-select': 'none'
+  'rtgl-svg':
+    'flex-shrink': '0'
```
han4wluc: if adding this to all `rtgl-svg` fixes it and does not cause any regressions. we should add it in the component itself: https://github.com/yuusoft-org/rettangoli/blob/85fa65a42459837ed0365ab43de30f8acf31e95c/packages/rettangoli-ui/src/primitives/svg.js#L17

src/components/baseFileExplorer/baseFileExplorer.view.yaml
```diff
@@ -77,6 +77,12 @@ styles:
...
+  'rtgl-text[ellipsis="true"]':
```
han4wluc: `flex` and `margin-right` can be achieved with this:

```
rtgl-view flex=1 mr=xl
```

`min-width` if we need, can add it to the component itself: https://github.com/yuusoft-org/rettangoli/blob/main/packages/rettangoli-ui/src/primitives/text.js#L86 (if need to apply only to ellipsis) or https://github.com/yuusoft-org/rettangoli/blob/main/packages/rettangoli-ui/src/primitives/text.js#L16 (if need to apply to all rtgl-text)

src/components/baseFileExplorer/baseFileExplorer.view.yaml
```diff
@@ -90,7 +96,7 @@ template:
                   - rtgl-view wh=16:
               - rtgl-view d=h g=md av=c w=f:
                 - rtgl-svg svg=${item.type} wh=16:
-                - rtgl-text s=sm ellipsis=true: "${item.name}"
+                - rtgl-text s=sm ellipsis=true w=f: "${item.name}"
```
han4wluc: prefer to write inline this way instead of using `styles`. use `styles` only if we need to apply it for more global stuff.

`rtgl-text s=sm ellipsis=true w=f flex=1 mr=xl style="min-width: 0"`

src/components/groupAudioView/groupAudioView.view.yaml
```diff
@@ -52,8 +52,7 @@ template:
                       - $for item in group.children:
                           - 'rtgl-view#audio-item-${item.id} cur=p style="${item.selectedStyle}"':
                             - rtgl-image br=md w=200 h=150 src=/public/project_logo_placeholder.png:
+                      - rvn-drag-drop#drag-drop-item-${group.id} .uploadText=uploadText .acceptedFileTypes=acceptedFileTypes:
```
han4wluc, line 55: Next time for change like this, can add a screenshot in the PR to make it easier to review.

I think is a big ugly, but it is more usable than before. so ok for now. in future would want to improve it more.

Screenshot 2025-07-01 at 16 48 26

src/components/baseFileExplorer/baseFileExplorer.view.yaml
738NGX: flex-shrink is not necessary, with styles on rtgl-text, there is no effect when i remove that

src/components/baseFileExplorer/baseFileExplorer.view.yaml
738NGX: completed in the next commit

han4wluc (approved):

PR #33, opened 2025-07-02
fix popover diaplay bug

738NGX:
image

Now it looks good.

This problem cannot be solved with adjust z-index. Finally i remove the 'position:sticky' style for those views. This should cause the problem(Overlaps other composited content). And there is no other extra display bugs without this style.

### DIFF
--- a/src/pages/animations/animations.view.yaml
+++ b/src/pages/animations/animations.view.yaml
@@ -18,7 +18,7 @@ refs:
 
 template:
   - rtgl-view d=h w=f h=f: 
-    - 'rtgl-view w=f h=f w=240 bwr=xs style="z-index:1;"':
+    - 'rtgl-view w=f h=f w=240 bwr=xs z=1':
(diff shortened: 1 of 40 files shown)


han4wluc:
yeah, but with this we don't have the sticky header feature any more. we do want to keep the sticky header feature.

src/pages/animations/animations.view.yaml
```diff
@@ -18,7 +18,7 @@ refs:
 
 template:
   - rtgl-view d=h w=f h=f: 
-    - rtgl-view w=f h=f w=240 bwr=xs:
+    - 'rtgl-view w=f h=f w=240 bwr=xs style="z-index:1;"':
```
han4wluc: `rtgl-view w=f h=f w=240 bwr=xs z=1` should work. https://github.com/yuusoft-org/rettangoli/blob/main/packages/rettangoli-ui/vt/specs/primitives/view/view-zindex-01.html

han4wluc (approved):

PR #38, opened 2025-07-04
add images full screen preview & dimensions display

738NGX:


src/components/groupImagesView/groupImagesView.handlers.js
```diff
@@ -85,6 +83,60 @@ export const handleImageItemClick = (e, deps) => {
...
+
+  if (fileId && httpClient) {
+    try {
+      // Get the actual image URL
+      const { url } = await httpClient.creator.getFileContent({ 
+        fileId: fileId, 
+        projectId: 'someprojectId' 
+      });
+
+      const overlay = document.createElement('div');
```
han4wluc: this is very wrong way to do this. you should use the framework way to do it with `.view.js` file

look for example at how Dialog is implemented: https://github.com/yuusoft-org/rettangoli/blob/main/packages/rettangoli-ui/src/components/dialog/dialog.view.yaml#L39

src/components/groupImagesView/groupImagesView.handlers.js
```diff
@@ -85,6 +83,60 @@ export const handleImageItemClick = (e, deps) => {
...
+  const { httpClient } = deps;
+  
+  // Get the fileId from the image item
+  const fileImageElement = e.currentTarget.querySelector('rvn-file-image');
+  const fileId = fileImageElement?.getAttribute('fileId');
+
+  if (fileId && httpClient) {
+    try {
+      // Get the actual image URL
+      const { url } = await httpClient.creator.getFileContent({
```
han4wluc: we already have the fileId. we just use `rvn-file-image` component. no need to make request to get url.

src/components/detailPanel/detailPanel.view.yaml
```diff
@@ -39,12 +42,12 @@ template:
...
                       - rtgl-view pos=rel:
                         - 'rtgl-view#editable-image-${i} cur=p style="position: relative;"':
-                          - rvn-file-image br=md w=${field.width} h=${field.height} fileId=${field.fileId}:
+                          - rvn-file-image br=md w=${field.width} h=${field.height} fileId=${field.fileId} key=${field.key}:
                           - 'rtgl-view pos=abs cor=full bgc=rgba(0,0,0,0.5) av=c ah=c style="opacity: 0; transition: opacity 0.2s;" h-style="opacity: 1;"':
                             - rtgl-svg svg=upload wh=24 c=wh:
                         - 'input#file-input-${i} type=file accept="image/*" style="display: none;"':
                     $else:
-                      - rvn-file-image br=md w=${field.width} h=${field.height} fileId=${field.fileId}:
+                      - rvn-file-image br=md w=${field.width} h=${field.height} fileId=${field.fileId} key=${field.key}:
```
han4wluc: i don't think you need `key=${field.key}` does it do anything? if you really need it, can just do `key=${field.fileId}`

src/pages/images/images.handlers.js
```diff
@@ -119,6 +145,10 @@ export const handleDetailPanelImageSelected = async (e, deps) => {
   }
   
   try {
+    // Get image dimensions before uploading
+    const dimensions = await getImageDimensions(file);
+    console.log('Image dimensions for replacement', file.name, ':', dimensions);
```
han4wluc: remove logs

src/pages/images/images.handlers.js
```diff
@@ -147,6 +177,8 @@ export const handleDetailPanelImageSelected = async (e, deps) => {
             name: file.name,
             fileType: file.type,
             fileSize: file.size,
+            width: dimensions?.width,
```
han4wluc: if dimensions width and height is not available, fail the whole thing. all images should have width and height.

for now just do `alert('no image width or height')`. in future we implment better notification UI.

src/components/groupImagesView/groupImagesView.handlers.js
738NGX: css styles text on rvn-file-image cannot work, which leads to image overflow, especially images height > width.
therefore, in this version i use the origin img instead.

han4wluc: we need to fix it in 
* https://github.com/yuusoft-org/routevn-creator-client/blob/main/src/components/fileImage/fileImage.view.yaml
* or even in * https://github.com/yuusoft-org/rettangoli/blob/main/packages/rettangoli-ui/src/primitives/image.js

src/components/groupImagesView/groupImagesView.handlers.js
```diff
@@ -85,6 +83,30 @@ export const handleImageItemClick = (e, deps) => {
...
 };
 
+export const handleImageDoubleClick = (e, deps) => {
+  const { store, render } = deps;
+  
+  // Get the fileId from the image item
+  const fileImageElement = e.currentTarget.querySelector('rvn-file-image');
+  const fileId = fileImageElement?.getAttribute('fileid');
+  
+  if (fileId) {
```
han4wluc: remove this check. fileId will always exist

src/components/groupImagesView/groupImagesView.handlers.js
```diff
@@ -85,6 +83,30 @@ export const handleImageItemClick = (e, deps) => {
   }));
 };
 
+export const handleImageDoubleClick = (e, deps) => {
+  const { store, render } = deps;
+  
+  // Get the fileId from the image item
+  const fileImageElement = e.currentTarget.querySelector('rvn-file-image');
```
han4wluc: do like this:

`const fileId = e.currentTarget.id.replace('rvn-file-image', '')`

because

`querySelector` is very unreliabile. it will easily break of the html structure is changed.

src/pages/images/images.handlers.js
```diff
@@ -32,6 +51,10 @@ export const handleDragDropFileSelected = async (e, deps) => {
   // Create upload promises for all files
   const uploadPromises = Array.from(files).map(async (file) => {
     try {
+      // Get image dimensions before uploading
+      const dimensions = await getImageDimensions(file);
```
han4wluc, line 55: if dimensions not avialble, add this:

```
        return {
          success: false,
          file,
          error: response.statusText,
        };
```

src/components/groupImagesView/groupImagesView.store.js
```diff
@@ -28,6 +30,17 @@ export const setZoomLevel = (state, zoomLevel) => {
   state.zoomLevel = newZoomLevel;
 }
 
+export const setPreviewFileId = (state, fileId) => {
```
han4wluc: better naming:


```
export const showFullImagePreview = (state, fileId) => {

}

// no need pass function args here. will always set to false and undefined.
export const hideFullImagePreview = (state) => {

}

```

also update store state property name accordingly

src/components/groupImagesView/groupImagesView.store.js
```diff
@@ -2,6 +2,8 @@ export const INITIAL_STATE = Object.freeze({
   collapsedIds: [],
   searchQuery: '',
   zoomLevel: 1.0,
+  previewActivated: false,
+  previewFileId: null,
```
han4wluc: we have anunwritten convetion we have. use `undefined` instead of `null`

src/pages/images/images.handlers.js
han4wluc: need to be safe here. because uploaded files can be unpredicatable.
if bad file is uploaded we will be storing bad data which can be problematic.

src/components/groupImagesView/groupImagesView.view.yaml
```diff
@@ -96,4 +106,9 @@ template:
           - rtgl-view w=f h=400 av=c ah=c:
             - rtgl-text s=lg c=mu-fg: No images found matching "${searchQuery}"
     - rtgl-view h=33vh:
+  - $if previewActivated:
+      - rtgl-view#preview-overlay pos=fix cor=full ah=c av=c z=3000 cur=p:
+        - rtgl-view w=f h=f pos=fix cor=full ah=c av=c bgc=bg op="0.8":
```
han4wluc: `op="0.5"` to be consistent with the other overlays we have.

src/components/groupImagesView/groupImagesView.handlers.js
738NGX: `const fileId = e.currentTarget.id.replace('rvn-file-image', '')` such a fileId is different from `querySelector` got. This will cause a 404 error.

han4wluc: sorry should have been `.replace('image-item', '')` need to look the `id` of the element we attached the event

src/pages/images/images.store.js
```diff
@@ -45,6 +45,7 @@ export const toViewData = ({ state, props }, payload) => {
     displayFileType: selectedItem.fileType || (selectedItem.type === 'image' ? 'PNG' : null),
     displayFileSize: selectedItem.fileSize ? formatFileSize(selectedItem.fileSize) : null,
     fullPath: selectedItem.fullLabel || selectedItem.name || '',
+    dimensions: (selectedItem.width && selectedItem.height) ? { width: selectedItem.width, height: selectedItem.height } : null,
```
han4wluc: here we just use `{ width: selectedItem.width, height: selectedItem.height }`
we assume is awlays avaialbe, since we already check that it will always be available during uplaod.

src/components/groupImagesView/groupImagesView.handlers.js
738NGX: for example,
`querySelector` got `2XVVYhPaGooCkNs9vJaj5fWv`
`replace` got `-WMLg0iLiFJWwYHBtCkCiV`
they are totally different

han4wluc: oh, I see.
one is itemId
one is item.fileId

the correct way then is to:

from itemId you get item. then from item you get item.fileId

src/components/groupImagesView/groupImagesView.store.js
```diff
@@ -28,6 +30,16 @@ export const setZoomLevel = (state, zoomLevel) => {
   state.zoomLevel = newZoomLevel;
 }
 
+export const showFullImagePreview = (state, fileId) => {
+  state.previewActivated = true
```
han4wluc: `fullImagePreviewVisible`
`fullImagePreviewFileId`

han4wluc (approved):

PR #43, opened 2025-07-07
Add Rename in Detail Pannel

738NGX:


src/components/detailPanel/detailPanel.handlers.js
```diff
@@ -40,4 +40,61 @@ export const handleFileInputChange = (e, deps) => {
...
+  render();
+};
+
+export const handleFormActionClick = (e, deps) => {
+  const { store, dispatchEvent, render } = deps;
+  const detail = e.detail;
+  
+  // Extract action and values from detail
+  const action = detail.action || detail.actionId;
+  const values = detail.values || detail.formValues || detail;
```
han4wluc: there should be only 1 correct. let's use the correct one without fallback.
```
  const action = detail.action || detail.actionId;
  const values = detail.values || detail.formValues || detail;
```

src/components/detailPanel/detailPanel.handlers.js
```diff
@@ -40,4 +40,61 @@ export const handleFileInputChange = (e, deps) => {
...
+    return;
+  }
+  
+  if (action === 'submit') {
+    // Hide popover
+    store.hidePopover();
+    render();
+    
+    // Emit file-action event for rename confirmation
+    dispatchEvent(new CustomEvent("file-action", {
```
han4wluc, line 91:
```
dispatchEvent(new CustomEvent("update-item-property", {
      detail: {
        itemProperty: 'name',
        value: values.name
      },
      bubbles: true,
      composed: true
    }));
  }
```

han4wluc: make it more generalized. so in future we can easily update other fields like descriptioon.

src/components/detailPanel/detailPanel.handlers.js
738NGX: now use
```
const action = detail.actionId;
const values = detail.formValues;
```

src/components/detailPanel/detailPanel.handlers.js
738NGX: have renamed all `file-action` or `handleFileAction` to `update-item-property`, `handleUpdateItemProperty`

src/components/baseFileExplorer/baseFileExplorer.handlers.js
```diff
@@ -329,8 +329,8 @@ export const handleFormActionClick = (e, deps) => {
     store.hidePopover();
     render();
     
-    // Emit file-action event for rename confirmation
-    dispatchEvent(new CustomEvent("file-action", {
```
han4wluc: oh, I see now. previous changes for detail panel was following the implementation of file explorer?
if that is the case. then yes, let's try be consistent and follow implmentation of file explorer. let's keep `file-action` at least is consistent.
maybe in future I will rewrite both to support fields other than name.

738NGX: well, now is all back to `file-action`

han4wluc (approved):

PR #45, opened 2025-07-08
Fix sprites page render bug

738NGX:
Now when pressing Sprites, it should render same stuff as `images`

### DIFF
--- a/src/deps/repository.js
+++ b/src/deps/repository.js
@@ -95,8 +95,17 @@ const removeNodeFromTree = (tree, nodeId) => {
 // Tree manipulation functions
 const treePush = (state, target, value) => {
   const newState = structuredClone(state);
-  const targetData = get(newState, target);
+  let targetData = get(newState, target);
   const { parent, item, position } = value;
+  
+  // If targetData is undefined, the path doesn't exist - create it
+  if (!targetData) {
+    console.warn(`treePush: Target path '${target}' does not exist, creating it`);
+    targetData = { tree: [], items: {} };
+    // Use set to create the path and get updated state
+    const stateWithPath = set(newState, target, targetData);
+    return treePush(stateWithPath, target, value);
+  }
   
   // Ensure tree and items exist
   if (!targetData.tree) {
--- a/src/deps/repository.js
+++ b/src/deps/repository.js
@@ -95,24 +95,8 @@ const removeNodeFromTree = (tree, nodeId) => {
 // Tree manipulation functions
 const treePush = (state, target, value) => {
   const newState = structuredClone(state);
-  let targetData = get(newState, target);
+  const targetData = get(newState, target);
   const { parent, item, position } = value;
-  
-  // If targetData is undefined, the path doesn't exist
-  if (!targetData) {
-    // Special case: if this is a character sprites path, create it (for backwards compatibility)
-    if (target.includes('characters.items.') && target.endsWith('.sprites')) {
-      const characterId = target.split('.')[2];
-      const characters = get(newState, 'characters');
-      if (characters && characters.items && characters.items[characterId]) {
-        // Character exists but doesn't have sprites - create it
-        targetData = { tree: [], items: {} };
-        const stateWithPath = set(newState, target, targetData);
-        return treePush(stateWithPath, target, value);
-      }
-    }
-    throw new Error(`Target path '${target}' does not exist. Cannot push to non-existent path.`);
-  }
   
   // Ensure tree and items exist
   if (!targetData.tree) {
--- a/src/deps/repository.js
+++ b/src/deps/repository.js
@@ -100,7 +100,6 @@ const treePush = (state, target, value) => {
   
   // If targetData is undefined, the path doesn't exist - create it
   if (!targetData) {
-    console.warn(`treePush: Target path '${target}' does not exist, creating it`);
     targetData = { tree: [], items: {} };
     // Use set to create the path and get updated state
     const stateWithPath = set(newState, target, targetData);
--- a/src/deps/repository.js
+++ b/src/deps/repository.js
@@ -98,12 +98,20 @@ const treePush = (state, target, value) => {
   let targetData = get(newState, target);
   const { parent, item, position } = value;
   
-  // If targetData is undefined, the path doesn't exist - create it
+  // If targetData is undefined, the path doesn't exist
   if (!targetData) {
-    targetData = { tree: [], items: {} };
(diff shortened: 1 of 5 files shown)


src/pages/characterSprites/characterSprites.handlers.js
```diff
@@ -2,24 +2,23 @@
...
-  const { router } = deps;
+  const { router, store, repository } = deps;
   const { characterId } = router.getPayload();
-  const { store, repository } = deps;
   const { characters } = repository.getState();
   const character = characters.items[characterId];
+  
   store.setCharacterId(characterId);
-  store.setItems(character.sprites);
+  store.setItems(character?.sprites || { tree: [], items: {} });
```
han4wluc: we don't want to do this.

if user goes to url `/project/resources/character-sprites?characterId=somerandomid`, then this record will be created and stored. but the charcterId does not exist.

character should be created first. when user creates the character we should call the repository to create characrter with empty sprites. that is the only entrypoint to create the charcter

src/deps/repository.js
```diff
@@ -95,9 +95,17 @@ const removeNodeFromTree = (tree, nodeId) => {
 // Tree manipulation functions
 const treePush = (state, target, value) => {
   const newState = structuredClone(state);
-  const targetData = get(newState, target);
+  let targetData = get(newState, target);
   const { parent, item, position } = value;
   
+  // If targetData is undefined, the path doesn't exist - create it
+  if (!targetData) {
```
han4wluc: this is more convenient, but there is risk of accidentally creating data in the wrong path.
is safer to throw an error if targetData does not exist.

src/pages/characterSprites/characterSprites.handlers.js
```diff
@@ -2,24 +2,23 @@
...
 
 export const handleDataChanged = (e, deps) => {
-  const { router, render } = deps;
+  const { router, render, store, repository } = deps;
   const { characterId } = router.getPayload();
-  const { store, repository } = deps;
   const { characters } = repository.getState();
   const character = characters.items[characterId];
-  store.setItems(character.sprites);
+  store.setItems(character?.sprites || { tree: [], items: {} });
```
han4wluc: character?.sprites should always exist and created properly. so no need to fallback.

src/pages/characterSprites/characterSprites.handlers.js
han4wluc: for now if character does not exist. can just do a
`alert("character not found")` or something.

src/pages/characterSprites/characterSprites.store.js
```diff
@@ -34,11 +34,28 @@ export const selectCharacterId = ({ state }) => {
...
   if (!state.selectedItemId) return null;
   // state.spritesData contains the full structure with tree and items
+  if (!state.spritesData || !state.spritesData.items || !state.spritesData.tree) return null;
   const flatItems = toFlatItems(state.spritesData);
   return flatItems.find(item => item.id === state.selectedItemId);
 }
 
 export const toViewData = ({ state, props }, payload) => {
+  // Check if spritesData has the expected structure
+  if (!state.spritesData || !state.spritesData.items || !state.spritesData.tree) {
```
han4wluc: based on comments above. this should not be needed anymore

src/pages/characterSprites/characterSprites.handlers.js
```diff
@@ -142,8 +143,9 @@ export const handleFileAction = (e, deps) => {
     });
     
     // Update the store with the new repository state
-    const { characterSprites } = repository.getState();
-    store.setItems(characterSprites);
+    const { characters } = repository.getState();
+    const character = characters.items[store.selectCharacterId()];
+    store.setItems(character?.sprites || { tree: [], items: {} });
```
han4wluc: based on comments above. this should not be needed anymore

src/pages/characterSprites/characterSprites.handlers.js
```diff
@@ -84,30 +83,32 @@ export const handleDragDropFileSelected = async (e, deps) => {
...
-
-  if (successfulUploads.length > 0) {
-    const { store, repository } = deps;
+    
+    // Update store with the latest repository state
     const { characters } = repository.getState();
-    const character = characters.items[store.selectCharacterId()];
-    store.setItems(character.sprites);
+    const character = characters.items[characterId];
+    store.setItems(character?.sprites || { tree: [], items: {} });
```
han4wluc: based on comments above. this should not be needed anymore

src/pages/characterSprites/characterSprites.handlers.js
```diff
@@ -84,30 +83,32 @@ export const handleDragDropFileSelected = async (e, deps) => {
...
-        parent: id,
-        position: "last",
-        item: {
-          id: nanoid(),
-          type: "image",
-          fileId: result.fileId,
-          name: result.file.name,
-          fileType: result.file.type,
-          fileSize: result.file.size,
+  if (successfulUploads.length > 0) {
```
han4wluc, line 106: based on comments above. this should not be needed anymore

src/pages/characterSprites/characterSprites.handlers.js
738NGX: simply raise error for all case will cause the tree creation failure, and the page won't rander.

```
repository.js:114 
 Uncaught Error: Target path 'characters.items.3A1kpT2EgReXA8JZHrKua.sprites' does not exist. Cannot push to non-existent path.
    at treePush (repository.js:114:11)
    at repository.js:359:16
    at Array.reduce (<anonymous>)
    at Object.getState (repository.js:352:25)
    at Object.handleOnMount28 (project.handlers.js:5:28)
```

i don't know which characters is it, because i have cleared the character list, but the error still exists.

Therefore, the temp solution in the commit is check if the character id exists, if true, it will create sth. to avoid error, or it will raise error and alert character not found.

738NGX: the problems is from browser local storage.
now removed all changes on repository.js and it still works well.

src/pages/characterSprites/characterSprites.store.js
han4wluc: @738NGX this is not needed anymore right?

han4wluc (approved):

src/pages/characterSprites/characterSprites.store.js
738NGX: well now removed, it can work

PR #50, opened 2025-07-09
Add Audio Player

738NGX:
It seems hard to implement directly in `src/components/groupAudioView`, as a solution, i rewrite the ui of `src/components/audioPlayer`, which before uses in detailPannel and now have been removed, and then put it to that position in `groupAudioView`.

Still have problems on changing audios, will try to solve tomorrow.

### DIFF
--- a/src/components/groupAudioView/groupAudioView.store.js
+++ b/src/components/groupAudioView/groupAudioView.store.js
@@ -21,9 +21,11 @@ export const toggleGroupCollapse = (state, groupId) => {
   }
 }
 
-export const openAudioPlayer = (state, fileId) => {
+export const openAudioPlayer = (state, { fileId, fileName }) => {
   state.playingAudio.fileId = fileId;
+  state.playingAudio.title = fileName || 'Audio File';
   state.showAudioPlayer = true;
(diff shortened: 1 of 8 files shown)


src/components/groupAudioView/groupAudioView.store.js
```diff
@@ -1,5 +1,12 @@
+import { current } from "immer";
```
han4wluc: I think this was unintentinal import

src/pages/audio/audio.store.js
```diff
@@ -41,7 +41,7 @@ export const toViewData = ({ state, props }, payload) => {
   // Transform selectedItem into detailPanel props
   const detailTitle = selectedItemDetails ? 'Audio Details' : null;
   const detailFields = selectedItemDetails && selectedItemDetails.type === 'audio' ? [
-    { type: 'audio', fileId: selectedItemDetails.fileId, autoPlay: true },
+    { type: 'audio', fileId: selectedItemDetails.fileId, fileName: selectedItemDetails.name, autoPlay: true },
```
han4wluc: we are removing audio from details panel. will just show an image instead.

738NGX: now totally removed

PR #53, opened 2025-07-10
refactor file uploader

han4wluc:
- **update**
- **update**


src/deps/createFileUploader.js
```diff
@@ -0,0 +1,78 @@
...
+    img.onerror = () => {
+      URL.revokeObjectURL(url); // Clean up memory
+      resolve(null);
+    };
+
+    img.src = url;
+  });
+};
+
+export const createFileUploader = ({ httpClient }) => {
```
738NGX: this function seems only work for image files. but the function name is general.

maybe a specific name like `createImageFileUploader` is more suitable?

han4wluc:
@738NGX updated. please review again

738NGX (approved):
ok

PR #64, opened 2025-07-11
add basic form for entering and storing typography info

aayushkdev:


src/components/detailPanel/detailPanel.store.js
```diff
@@ -31,6 +31,66 @@ export const INITIAL_STATE = Object.freeze({
...
+        id: 'fontSize',
+        fieldName: 'fontSize',
+        inputType: 'inputText',
+        label: 'Font Size',
+        description: 'Enter the font size (e.g., 16, 18, 24)',
+        required: true,
+      }, {
+        id: 'fontColor',
+        fieldName: 'fontColor',
+        inputType: 'inputText',
```
han4wluc: for fontColor, fontStyle should use select.

https://github.com/yuusoft-org/rettangoli/blob/766edf4914100acc3ffabf1e38e14026d878c098/packages/rettangoli-ui/vt/specs/components/form/mixed-inputs.html#L49

the value that we store should be the color.id, and fontStyle.id. what we show to user is the name.

src/pages/typography/typography.handlers.js
```diff
@@ -133,3 +135,103 @@ export const handleTypographyCreated = (e, deps) => {
...
+  if (!name || !fontSize || !fontColor || !fontStyle || !fontWeight) {
+    alert('Please fill in all required fields');
+    return;
+  }
+  
+  // Get repository data for validation
+  const { fonts, colors } = repository.getState();
+  
+  // Check if color exists
+  const colorExists = toFlatItems(colors)
```
han4wluc: if we use the `select` from existing list. there shoul be no need to do all those checks.

src/components/detailPanel/detailPanel.store.js
```diff
@@ -68,6 +124,40 @@ export const hideColorDialog = (state) => {
...
+  state.typographyDialog.defaultValues = {
+    fontSize: itemData.fontSize || '16',
+    fontColor: itemData.fontColor || '#000000',
+    fontStyle: itemData.fontStyle || '',
+    fontWeight: itemData.fontWeight || 'normal',
+  };
+  
+  // Update dropdown options dynamically
+  if (colorOptions) {
+    state.typographyDialog.form.fields[1].options = colorOptions;
```
han4wluc: from programming perspective hardocdng the array index is not reliable. the order could change, and then would cause unintetional bug.

better to do 
```
state.typographyDialog.for.fields.find(id => id == 'fontColor').options = colorOptions.
```

han4wluc (approved):

han4wluc (approved):

PR #68, opened 2025-07-13
add resizable left file explorer panel

aayushkdev:


src/components/resizablePanel/resizablePanel.store.js
```diff
@@ -0,0 +1,63 @@
+// Load panel width from localStorage, fallback to 280
+const getStoredPanelWidth = () => {
+  try {
+    const stored = localStorage.getItem('resizablePanelWidth');
```
han4wluc: this is very wrong. should use this instead:  https://github.com/yuusoft-org/routevn-creator-client/blob/main/src/deps/userConfig.js

src/components/resizablePanel/resizablePanel.view.yaml
```diff
@@ -0,0 +1,44 @@
+elementName: rvn-resizable-panel
```
han4wluc, line 1: this works, but I would like to explore other options of impllenting this.
with this we need to add 1 more layer of component. which I would like to minimize.

aayushkdev:
Ready for final review

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -0,0 +1,82 @@
...
+  const { store, attrs, userConfig } = deps;
+
+  const panelType = attrs.panelType || "file-explorer";
+
+  // Set different default widths based on panel type
+  let defaultWidth;
+  if (panelType === "detail-panel") {
+    defaultWidth = parseInt(attrs.w) || 270; // Default width for detail panels
+  } else {
+    defaultWidth = parseInt(attrs.w) || 280; // Normal width for file explorer
```
han4wluc: we should not hardcode with for detail panel or file explorer. just choose 280.

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export const handleResizeStart = (e, deps) => {
+  const { store, render, attrs } = deps;
+
+  e.preventDefault();
+  console.log("🔧 Resizable panel resize start triggered");
+
+  const startX = e.clientX;
+  const startWidth = store.getState().panelWidth;
```
han4wluc: don't use getState()

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -0,0 +1,82 @@
...
+    defaultWidth = parseInt(attrs.w) || 280; // Normal width for file explorer
+  }
+
+  // Load from localStorage using userConfig pattern
+  const configKey = `resizablePanel.${panelType}Width`;
+  const storedWidth = userConfig.get(configKey);
+  const width = storedWidth ? parseInt(storedWidth, 10) : defaultWidth;
+
+  store.initializePanelWidth(attrs);
+  store.setPanelWidth(width, attrs);
```
han4wluc: we don't need 2 store actions here. only have 1.
also don't pass whole attrs into store action. only filter the values that are actually needed or used.

```
store.initializePanelWidth({
  width,
})
```

src/components/resizablePanel/resizablePanel.store.js
```diff
@@ -0,0 +1,50 @@
...
+  state.startX = startX;
+  state.startWidth = startWidth;
+};
+
+export const initializePanelWidth = (state, attrs) => {
+  const panelType = attrs.panelType || "file-explorer";
+
+  // Set different default widths based on panel type
+  let defaultWidth;
+  if (panelType === "detail-panel") {
```
han4wluc: default is already handled from outside. no need to handle it again here.

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -0,0 +1,82 @@
+export const handleBeforeMount = (deps) => {
+  const { store, attrs, userConfig } = deps;
+
+  const panelType = attrs.panelType || "file-explorer";
```
han4wluc: `attrs['panel-type']`

attribute should always be kebab case

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -0,0 +1,82 @@
...
+  const { store, render, attrs } = deps;
+
+  e.preventDefault();
+  console.log("🔧 Resizable panel resize start triggered");
+
+  const startX = e.clientX;
+  const startWidth = store.getState().panelWidth;
+
+  store.setIsResizing(true);
+  store.setResizeStart({ startX, startWidth });
```
han4wluc: merge `setIsResizing` and `setResizeStart` into one action.

src/components/resizablePanel/resizablePanel.store.js
```diff
@@ -0,0 +1,50 @@
...
+    defaultWidth = parseInt(attrs.w) || 270; // Default width for detail panels
+  } else {
+    defaultWidth = parseInt(attrs.w) || 280; // Normal width for file explorer
+  }
+
+  state.panelWidth = defaultWidth;
+  state.startWidth = defaultWidth;
```
han4wluc: need to apply min and max here as well.

aayushkdev:
I have added the changes

han4wluc (approved):
LGTM

PR #86, opened 2025-07-18
update placement properties

han4wluc:
Replaced with new properties from `route-graphics`

which follows this format:

```
    state.defaultValues = {
      name: "",
      x: "0",
      y: "0",
      scaleX: "1",
      scaleY: "1",
      anchor: { anchorX: 0, anchorY: 0 },
      rotation: "0",
    };
```

There is an existing but which defaultValue don't work for anchor selector because value is an object. 

Screenshot 2025-07-18 at 16 41 45


738NGX (approved):
Anchor may both support selector and slider-input for custom settings, like unity, will consider to support later

PR #93, opened 2025-07-19
fix search bar not working

aayushkdev:


src/components/groupAnimationsView/groupAnimationsView.view.yaml
```diff
@@ -19,7 +19,7 @@ propsSchema:
 refs:
   search-input:
     eventListeners:
-      input-keydown:
+      input:
```
han4wluc: ok. i checked on this one.

input is emitted from the original `<input>` inside `rtgl-input`, this is unintended.

the correct event name is `input-change`. refer to: https://github.com/yuusoft-org/rettangoli/blob/14c0c97a34b7867254adbbdc5352a0f9e31431d3/packages/rettangoli-ui/src/primitives/input.js#L114


the payload will be in `e.detail.value`

I'll add e.stopPropagation() to `rtgl-input` so that `input` event will not be emitted.

aayushkdev:
I noticed that animations and audio didnt have search implemented so I have also added search there too along with the suggested changes

han4wluc (approved):
LGTM

PR #97, opened 2025-07-21
Zoom slider update

738NGX:


src/setup.js
```diff
@@ -127,6 +128,9 @@ const componentDependencies = {
   drenderer,
 };
 
+// Add zoom handlers factory - will be created at runtime with proper deps
+componentDependencies.createZoomHandlers = createZoomHandlers;
```
han4wluc: move this to line 129 insdie the object. also need to add to pageDependences. for now we keep them 2 the same.

src/components/groupAudioView/groupAudioView.view.yaml
```diff
@@ -1,10 +1,16 @@
 elementName: rvn-group-audio-view
 
-viewDataSchema:
+handleBeforeMountviewDataSchema:
```
han4wluc: this is wrong.

src/deps/zoomManager.js
```diff
@@ -0,0 +1,52 @@
+export const createZoomHandlers = (componentName, deps) => {
```
han4wluc, line 1: this pattern is weird. not ideal. it is too specific to this.
let me think of a better way to make reusable logic.

PR #104, opened 2025-07-23
add remaining features in animations page

aayushkdev:


han4wluc:
This PR is ok, can merge. 


the below still need to be done.

* move left should not appear if right click on 1st frame
* move right should not appear if right click on last frame
* try add default value for edit form, let me know if you have issues with it

han4wluc (approved):

PR #110, opened 2025-07-25
Layout feat 1

738NGX:
Known bugs or need to change:

- Container
  - Remove anchor
  - Test direction
  - Solve children offset problems in some situation (For example, mult sprites)
- Sprite 
  - Scale not work
  - Support hover and click change
  - More reasonable dimension control. Support auto size, or let user select image first when create sprite object.
- Text
  - Font, color, scale, anchor... many things need to do

image


src/components/detailPanel/detailPanel.handlers.js
```diff
@@ -30,6 +30,54 @@ export const handleSelectChange = (e, deps) => {
...
+export const handleEditableNumberClick = (e, deps) => {
+  const { store, render } = deps;
+  e.preventDefault();
+
+  const fieldName = e.currentTarget.id.replace("editable-number-", "");
+  const field = store.selectField(fieldName);
+
+  // Calculate position for left-bottom placement relative to mouse cursor
+  const position = {
+    x: e.clientX - 200, // Move 200px to the left of cursor
```
han4wluc: I think no need to use this. no need to calculate manually.. you should use `placement` instead.

src/components/detailPanel/detailPanel.handlers.js
```diff
@@ -30,6 +30,54 @@ export const handleSelectChange = (e, deps) => {
...
+    y: e.clientY + 10, // Move 10px down from cursor
+  };
+
+  // Ensure popover doesn't go off-screen to the left
+  if (position.x < 10) {
```
han4wluc: this is also automatically taken care of with placement.

src/pages/layoutEditor/layoutEditor.handlers.js
```diff
@@ -150,21 +234,31 @@ export const handleImageSelectorUpdated = async (e, deps) => {
...
 
-  const { layouts } = repository.getState();
+  const { layouts, images } = repository.getState();
   const layout = layouts.items[layoutId];
+
+  // Preserve selectedItemId before updating store
+  const currentSelectedItemId = store.selectSelectedItemId();
+
   store.setItems(layout?.elements || { items: {}, tree: [] });
+  store.setImages(images);
```
han4wluc: I think this is only needed to set once during `handleBeforeMount`

src/pages/layoutEditor/layoutEditor.handlers.js
```diff
@@ -150,21 +234,31 @@ export const handleImageSelectorUpdated = async (e, deps) => {
...
     value: {
-      id: store.selectSelectedItemId(),
+      id: selectedItemId,
       replace: false,
       item: { imageId },
     },
   });
 
-  const { layouts } = repository.getState();
+  const { layouts, images } = repository.getState();
```
han4wluc: convert to not use getState

han4wluc (approved):
LGTM. minor comments.

src/components/detailPanel/detailPanel.handlers.js
738NGX: the placement seems have offset problems, need solve later

PR #113, opened 2025-07-25
Add sprite hover and click feature

738NGX:
https://github.com/user-attachments/assets/7b6898fa-b50c-4c32-9542-4b8745c9a8f2

known todo:
- should have a way to remove selected hover and click image if user no longer want them
- should have a selected image preview in detail pannel, which is more obvious

### DIFF
--- a/src/pages/layoutEditor/layoutEditor.handlers.js
+++ b/src/pages/layoutEditor/layoutEditor.handlers.js
@@ -236,10 +236,8 @@ export const handleImageSelectorUpdated = async (e, deps) => {
   const layoutId = store.selectLayoutId();
   const selectedItemId = store.selectSelectedItemId();
 
-  // Get the field name from the fields array
-  const viewData = store.toViewData();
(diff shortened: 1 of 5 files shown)


src/pages/layoutEditor/layoutEditor.handlers.js
```diff
@@ -232,18 +232,23 @@ export const handleRequestImageGroups = (e, deps) => {
...
 export const handleImageSelectorUpdated = async (e, deps) => {
   const { repository, store, render } = deps;
-  const { imageId } = e.detail;
+  const { imageId, fieldIndex } = e.detail;
   const layoutId = store.selectLayoutId();
   const selectedItemId = store.selectSelectedItemId();
 
-  // Update the selected item with the new imageId
+  // Get the field name from the fields array
+  const viewData = store.toViewData();
```
han4wluc: dont use toViewData from habdler. it is supposed to be used for view only. create new selectors that you need.

han4wluc (approved):

PR #114, opened 2025-07-25
Support Image preview and clear

738NGX:
image

imgae clear use a button in detail pannel. temp solution


han4wluc (approved):
ok.
but the 'clear' button is a bit ugly. correct interaction should be right click to show a dropdown-menu, and click delete (same right click file explorer item)

PR #140, opened 2025-07-31
Fix font loading

738NGX:
Known Problems：
1. Now popover-input at layout editor cannot use, will report `rettangoli-iife-ui.min.js:919  Uncaught ReferenceError: props is not defined`
2. It will have times of black screen when re-rendering, may due to font loading


han4wluc (approved):
> Now popover-input at layout editor cannot use, will report rettangoli-iife-ui.min.js:919  Uncaught ReferenceError: props is not defined

I encountered this as well just now. will fix shortly.

PR #143, opened 2025-07-31
Feat layout type

738NGX:
image


src/pages/layouts/layouts.store.js
```diff
@@ -47,9 +52,15 @@ export const toViewData = ({ state, props }, payload) => {
   let defaultValues = {};
 
   if (selectedItem) {
+    const layoutTypeLabel = selectedItem.layoutType
+      ? selectedItem.layoutType.charAt(0).toUpperCase() +
```
han4wluc: better hardcoded object with the mapping rather than doing transformation.

han4wluc (approved):

PR #144, opened 2025-08-01
Feat text var

738NGX:
image
image
image

### DIFF
--- a/src/pages/layoutEditor/layoutEditor.store.js
+++ b/src/pages/layoutEditor/layoutEditor.store.js
@@ -399,17 +399,22 @@ export const toViewData = ({ state, props }, payload) => {
           ...(selectedItem.type === "text"
             ? [
                 {
-                  name: 'contentType',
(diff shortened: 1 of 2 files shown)


src/pages/layoutEditor/layoutEditor.store.js
```diff
@@ -397,9 +399,31 @@ export const toViewData = ({ state, props }, payload) => {
...
                 {
-                  name: "text",
-                  inputType: "popover-input",
-                  description: "Text Content",
+                  name: "contentType",
+                  inputType: "select",
+                  description: "Content Type",
+                  options: [
+                    { label: "Custom Content", value: "custom" },
+                    ...(state.layoutType === "dialogue"
```
han4wluc: this is ok. but I don't like to write like this. see if can be simpliefied.
ideally by checking `layoutType` from context

han4wluc (approved):

PR #154, opened 2025-08-06
Bugs fix

738NGX:


han4wluc:
This solution we duplicate too much stuff. I don't like it.
Is it possible to use pos=fix and then you get the left & right from the user config?

738NGX:
> This solution we duplicate too much stuff. I don't like it. Is it possible to use pos=fix and then you get the left & right from the user config?

it seems can not immediately update, because user config only update after mouse release

han4wluc:
> > This solution we duplicate too much stuff. I don't like it. Is it possible to use pos=fix and then you get the left & right from the user config?
> 
> it seems can not immediately update, because user config only update after mouse release

I see.
what we can do is to dispatch an event using `subject` from the resize panel for mouse move. then in the audio page, you subscribe to this event, and update the position.

738NGX:
> > > This solution we duplicate too much stuff. I don't like it. Is it possible to use pos=fix and then you get the left & right from the user config?
> > 
> > 
> > it seems can not immediately update, because user config only update after mouse release
> 
> I see. what we can do is to dispatch an event using `subject` from the resize panel for mouse move. then in the audio page, you subscribe to this event, and update the position.

Completed

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -56,11 +56,20 @@ const handleResizeMove = (e, deps) => {
 
   store.setPanelWidth(newWidth, attrs);
 
+  // Dispatch resize event via subject for other components to listen
+  if (subject) {
```
han4wluc: subject is always availabe. no need to check

src/components/resizablePanel/resizablePanel.handlers.js
```diff
@@ -71,6 +80,15 @@ const handleResizeEnd = (e, deps, listeners) => {
   const currentWidth = store.selectPanelWidth();
   userConfig.set(configKey, currentWidth.toString());
 
+  // Dispatch resize-end event via subject
+  if (subject) {
```
han4wluc: subject is always availabe. no need to check

src/components/groupAudioView/groupAudioView.handlers.js
```diff
@@ -77,3 +77,66 @@ export const handleAudioPlayerClose = (e, deps) => {
...
+    });
+
+    // Store subscription for cleanup
+    store._resizeSubscription = subscription;
+  }
+};
+
+export const handleBeforeUnmount = (deps) => {
+  const { store } = deps;
+  if (store._resizeSubscription) {
```
han4wluc: it is wrong to store this in store.

src/components/groupAudioView/groupAudioView.handlers.js
```diff
@@ -77,3 +77,66 @@ export const handleAudioPlayerClose = (e, deps) => {
...
+    });
+    render();
+  };
+
+  // Initial position
+  updatePosition();
+
+  // Listen for panel resize events using subject
+  const { subject } = deps;
+  if (subject) {
```
han4wluc: should use it like this like you did before in sidebar

https://github.com/yuusoft-org/routevn-creator-client/blob/01be1634b4d8fb704ce9359da7e8d4c6860edf76/src/pages/sidebar/sidebar.handlers.js#L55

src/components/groupAudioView/groupAudioView.handlers.js
```diff
@@ -77,3 +77,66 @@ export const handleAudioPlayerClose = (e, deps) => {
...
+    userConfig.get("resizablePanel.file-explorerWidth") || "280",
+    10,
+  );
+  let currentRightWidth = parseInt(
+    userConfig.get("resizablePanel.detail-panelWidth") || "320",
+    10,
+  );
+
+  // Function to update position
+  const updatePosition = () => {
```
han4wluc: this function is doing too little. better not have this function at all.
do the stuff inside this function directly

src/components/groupAudioView/groupAudioView.handlers.js
```diff
@@ -77,3 +77,66 @@ export const handleAudioPlayerClose = (e, deps) => {
...
+
+  // Store current panel widths
+  let currentLeftWidth = parseInt(
+    userConfig.get("resizablePanel.file-explorerWidth") || "280",
+    10,
+  );
+  let currentRightWidth = parseInt(
+    userConfig.get("resizablePanel.detail-panelWidth") || "320",
+    10,
+  );
```
han4wluc: the 1st time to call this should be in handleBeforeMount.

src/components/groupAudioView/groupAudioView.handlers.js
```diff
@@ -77,3 +77,66 @@ export const handleAudioPlayerClose = (e, deps) => {
...
+        // Update from userConfig when resize ends (in case of any discrepancy)
+        currentLeftWidth = parseInt(
+          userConfig.get("resizablePanel.file-explorerWidth") || "280",
+          10,
+        );
+        currentRightWidth = parseInt(
+          userConfig.get("resizablePanel.detail-panelWidth") || "320",
+          10,
+        );
+        updatePosition();
```
han4wluc: no need to have
```
        currentLeftWidth = parseInt(
          userConfig.get("resizablePanel.file-explorerWidth") || "280",
          10,
        );
        currentRightWidth = parseInt(
          userConfig.get("resizablePanel.detail-panelWidth") || "320",
          10,
        );
```

for `panel-resize-end` as well, can use same logic as `panel-resize`. should be the same.

han4wluc (commented):
logic is correct.
but code writing style is wrong.

738NGX:
All fixed

han4wluc (approved):
LGTM

PR #161, opened 2025-08-11
Feat animation 3

738NGX:
Basically finished

https://github.com/user-attachments/assets/10541e32-f421-41e2-a2c6-18af274a4252

### DIFF
--- a/src/components/groupAnimationsView/groupAnimationsView.handlers.js
+++ b/src/components/groupAnimationsView/groupAnimationsView.handlers.js
@@ -66,7 +66,7 @@ export const handleAddAnimationClick = async (e, deps) => {
 
   // Open dialog for adding
   store.openDialog();
-  drenderer.render(resetState)
+  drenderer.render(resetState);
   render();
 };
 
@@ -100,7 +100,7 @@ export const handleAnimationItemDoubleClick = async (e, deps) => {
 
   if (itemData) {
     // Open dialog for editing
-    drenderer.render(resetState)
+    drenderer.render(resetState);
     store.openDialog({ editMode: true, itemId, itemData });
     render();
   }
--- a/src/components/groupAnimationsView/groupAnimationsView.handlers.js
+++ b/src/components/groupAnimationsView/groupAnimationsView.handlers.js
@@ -277,8 +277,8 @@ export const handleAddPropertyFormSubmit = (e, deps) => {
 
   // Get default value for the property if using existing value
   const defaultValues = {
-    x: 0,
-    y: 0,
+    x: 960,
+    y: 540,
     alpha: 1,
     scaleX: 1,
     scaleY: 1,
@@ -607,8 +607,8 @@ export const handleEditInitialValueFormSubmit = (e, deps) => {
 
   // Get default value for the property if using existing value
   const defaultValues = {
-    x: 0,
-    y: 0,
+    x: 960,
+    y: 540,
     alpha: 1,
     scaleX: 1,
     scaleY: 1,
--- a/src/components/groupAnimationsView/groupAnimationsView.handlers.js
+++ b/src/components/groupAnimationsView/groupAnimationsView.handlers.js
@@ -1,4 +1,13 @@
-import { nanoid } from "nanoid";
+export const handleAfterMount = async (deps) => {
+  const { store, drenderer, getRefIds } = deps;
+  const { canvas } = getRefIds();
+  if (canvas && canvas.elm && !store.selectIsDrendererInitialized()) {
+    await drenderer.init({
+      canvas: canvas.elm,
+    });
+    store.setDrendererInitialized(true);
+  }
+};
 
 // Constants for preview rendering
 const CANVAS_WIDTH = 1920;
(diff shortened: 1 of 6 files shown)


src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -1,5 +1,121 @@
...
+                  ? parseFloat(kf.duration) * 1000
+                  : parseFloat(kf.duration) || 1000,
+              value: value,
+              easing: kf.easing || "linear",
+              relative: kf.relative === true,
+            };
+          }),
+        };
+      }
+    }
```
han4wluc: lets store the data as much as possible directly as it is used so we don't need to do these transformations.
the more of these transformations we have. the more bugs we will have.

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -42,6 +158,19 @@ export const handleAddAnimationClick = (e, deps) => {
   // Open dialog for adding
   store.openDialog();
   render();
+
+  // Initialize drenderer after dialog is opened and canvas is in DOM
+  const { canvas } = getRefIds();
+  if (canvas && canvas.elm && !drenderer.initialized) {
+    await drenderer.init({
```
han4wluc: it should be intialized only once in `handleAfterMount`. there should be no other places to initialize it.
if is not initialized, disable the animation preview button so that it cant run on uninitialized canvas.

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -42,6 +158,19 @@ export const handleAddAnimationClick = (e, deps) => {
...
+      canvas: canvas.elm,
+    });
+    drenderer.initialized = true;
```
han4wluc: don't do this. this is not a functionality we expose from `drenderer`.
if you want store a state. store it in a store state.

`drenderer.initialized` changes it globally, for all places where `drenderer` is used including other pages.

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -1,5 +1,121 @@
...
 
+// Constants for preview rendering
+const CANVAS_WIDTH = 1920;
+const CANVAS_HEIGHT = 1080;
+const BG_COLOR = "#4a4a4a";
+const PREVIEW_RECT_WIDTH = 200;
+const PREVIEW_RECT_HEIGHT = 200;
+
+// Helper to create render state with animations
+const createAnimationRenderState = (
```
han4wluc: see if you can move this into a selector into the store.

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -76,6 +205,23 @@ export const handleAnimationItemDoubleClick = (e, deps) => {
     // Open dialog for editing
     store.openDialog({ editMode: true, itemId, itemData });
     render();
+
+    // Initialize drenderer after dialog is opened and canvas is in DOM
+    const { canvas } = getRefIds();
+    if (canvas && canvas.elm) {
+      await drenderer.init({
```
han4wluc: init should be done only in `handleAfterMount`

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -164,6 +310,14 @@ export const handleAddPropertyFormSubmit = (e, deps) => {
   store.addProperty({ property, initialValue: finalInitialValue });
   store.closePopover();
   render();
+
+  // Update preview
+  if (drenderer.initialized) {
+    const formState = store.selectFormState();
+    const animationProperties = formState.animationProperties || {};
+    const renderState = createAnimationRenderState(animationProperties, false);
+    drenderer.render(renderState);
```
han4wluc: do we really need to render this if user does not click the play animation button?

src/components/groupAnimationsView/groupAnimationsView.handlers.js
```diff
@@ -258,9 +420,29 @@ export const handleKeyframeDropdownItemClick = (e, deps) => {
...
     store.closePopover();
+    // Update preview after deleting property
+    if (drenderer.initialized) {
+      const formState = store.selectFormState();
+      const animationProperties = formState.animationProperties || {};
+      const renderState = createAnimationRenderState(
+        animationProperties,
+        false,
+      );
+      drenderer.render(renderState);
```
han4wluc: do we really need all these.
if need, there must be better way than wirting the same thing mulitple times.

src/components/groupAnimationsView/groupAnimationsView.store.js
```diff
@@ -448,7 +451,12 @@ export const updateKeyframe = (state, payload) => {
   console.log("state.animationProperties", state.animationProperties);
   const { property, index, keyframe } = payload;
   const keyframes = state.animationProperties[property].keyframes;
-  keyframes[index] = keyframe;
+
+  // Convert relative from string to boolean
+  keyframes[index] = {
+    ...keyframe,
+    relative: keyframe.relative === "true",
```
han4wluc: we should fix it to store boolean

738NGX:
All fixed

han4wluc (approved):

PR #170, opened 2025-08-13
feat layout anchor selector

738NGX:
image


src/pages/layoutEditor/layoutEditor.handlers.js
```diff
@@ -302,7 +302,19 @@ export const handleFormChange = async (e, deps) => {
   const layoutId = store.selectLayoutId();
   const selectedItemId = store.selectSelectedItemId();
 
-  const unflattenedUpdate = unflattenKey(e.detail.name, e.detail.fieldValue);
+  let unflattenedUpdate;
+
+  // Handle anchor selection specially
+  if (e.detail.name === "anchor" && e.detail.fieldValue) {
```
han4wluc: actually. value should be able to take an object. you can try that. but maybe have bug for default value or other things.

han4wluc (approved):

src/pages/layoutEditor/layoutEditor.handlers.js
738NGX: Yes, object value can work, the problem is the selector won't show the correct label

PR #171, opened 2025-08-13
Bugs fix 2

738NGX:
image

image

image


src/pages/sidebar/sidebar.store.js
```diff
@@ -1,3 +1,9 @@
+import {
+  assetItems,
+  userInterfaceItems,
+  systemConfigItems,
+} from "../resourceTypes/resourceTypes.store.js";
```
han4wluc, line 5: should not import from store to store. store should be completely independent of each other.

if really want to do this, create the conststss in a utils or common file. and import it from both stores.

han4wluc: better hard code this, or create the hardcoded config dynamically. example:

```
const selectMaps = {
  'assets': [
    "/project/resources/images", "/project/resources/audio", "..."
   ]
}
```

dynamically

generate selectMaps from assetItems, userInterfaceItems, systemConfigItems

han4wluc (approved):

PR #174, opened 2025-08-14
fix scene editor font load

738NGX:
image

### DIFF
--- a/src/pages/fonts/fonts.handlers.js
+++ b/src/pages/fonts/fonts.handlers.js
@@ -31,6 +31,15 @@ export const handleFontItemClick = (e, deps) => {
   render();
 };
 
+const getFileType = (result) => {
+  if (result.file.type) return result.file.type;
+  const ext = result.file.name.split(".").pop()?.toLowerCase();
+  if (result.type === "font" && ext) {
+    return `font/${ext}`;
+  }
+  throw new Error("Unknown file type");
+};
+
 export const handleFormExtraEvent = async (e, deps) => {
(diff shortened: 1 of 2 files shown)


src/pages/fonts/fonts.handlers.js
```diff
@@ -78,7 +78,10 @@ export const handleFormExtraEvent = async (e, deps) => {
         fileId: uploadResult.fileId,
         name: uploadResult.file.name,
         fontFamily: uploadResult.fontName,
-        fileType: uploadResult.file.type,
+        fileType:
+          uploadResult.type === "font"
+            ? `font/${uploadResult.file.name.split(".").pop()?.toLowerCase() || "ttf"}`
```
han4wluc: filename is not 100% reliable, I would use it only as last resort.
is there no value for `uploadResult.file.type`?
if the value is empty, seems more like a browser issue. would need to search online & AI to see if this is a known issue on the internet.

738NGX: uploadResult.type got "font", uploadResult.file.type got no value

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -17,6 +17,7 @@ async function createAssetsFromFileIds(
         projectId: "someprojectId",
       });
       let type;
+      let itemName;
```
han4wluc: remove this one if not used.

han4wluc (approved):

PR #186, opened 2025-08-18
Fix init template data

738NGX:


src/setup.js
```diff
@@ -207,122 +274,237 @@ if (!templateCreated) {
     templateFontsData.fetchedFonts,
   );
```
738NGX, line 273: Here can load other template if we want in the future

src/setup.js
```diff
@@ -207,122 +274,237 @@ if (!templateCreated) {
...
+          parent: "default-layouts-group",
+          item: { id, ...item },
+          position: "last",
+        },
+      });
+    }
+  }
+
+  // Immediately save to localStorage after adding all template data
+  localStorage.setItem(
```
han4wluc: implement `repository.flush()` which will immediately save to local storage instead of waiting for next save interval.

doing localStorage.setItem manually is too hacky. we should use official repository interface.

src/setup.js
```diff
@@ -207,122 +274,237 @@ if (!templateCreated) {
...
-    },
-  };
-} else {
-  // Template already created - still need to provide template structure for action stream
-  console.log("Template already exists - providing template structure...");
+  // Add template data using repository actions
+  // Add images - folders first, then images
+  for (const [id, item] of Object.entries(templateImagesData.imageItems)) {
+    if (item.type === "folder") {
+      repository.addAction({
```
han4wluc: the below is too repetitive.

create and implement a new action type instead:

```
repository.addAction('init', theEntireObject)
```

src/setup.js
```diff
@@ -188,14 +188,81 @@ async function fetchTemplateFonts() {
...
+  },
+};
 
-let initialData;
+const localStorageKey = "repositoryEventStream";
 
-if (!templateCreated) {
-  // First time - create everything
-  console.log("First time user - creating template data...");
+const repository = createRepository(initialData, localStorageKey);
```
han4wluc, line 256: if we use `repository.getActionStream().length`, there should be no need for `const templateCreated = localStorage.getItem("templateFilesUploaded");` anymore

src/setup.js
738NGX: Meet `Uncaught TypeError: repository.flush is not a function`

han4wluc: yes, it does not exist right now. you need to go into `repository.js` and implement that function.

738NGX:
Now all fixed

src/deps/repository.js
```diff
@@ -328,19 +328,27 @@ export const createRepository = (initialState, localStorageKey) => {
 
     const repository = createRepositoryInternal(initialState, actionStream);
 
-    // Auto-save to localStorage every 5 seconds
-    setInterval(() => {
+    // Add flush method to manually save to localStorage
+    repository.flush = () => {
```
han4wluc: we don't update instance functions.
should be like this:

```
const createRepositoryInternal = () => {
  ......

  return {
      .........,
      flush: () => {
           ...
      }
  }
}
```

han4wluc (approved):
good to merge after fixing the flush

PR #196, opened 2025-08-20
Feat color add entry in typography color selector

738NGX:
https://github.com/user-attachments/assets/482c076a-1113-4700-bb02-275691a10877

Need check if this implement is ok, decide how to implement others

### DIFF
--- a/src/pages/typography/typography.store.js
+++ b/src/pages/typography/typography.store.js
@@ -1,5 +1,45 @@
 import { toFlatGroups, toFlatItems } from "../../deps/repository";
 
+// Helper function to create add color form
+const createAddColorForm = (colorFolderOptions) => ({
+  title: "Add New Color",
+  description: "Create a new color for typography",
+  fields: [
+    {
+      name: "name",
+      inputType: "inputText",
+      label: "Color Name",
+      description: "Enter the color name",
+      required: true,
+    },
+    {
+      name: "hex",
+      inputType: "colorPicker",
+      label: "Hex Value",
+      description: "Choose or enter a hex color value",
+      required: true,
+    },
+    {
+      name: "folderId",
+      inputType: "select",
+      label: "Folder",
+      description: "Choose where to save the color",
+      options: colorFolderOptions,
+      required: true,
+    },
+  ],
+  actions: {
+    layout: "",
+    buttons: [
+      {
+        id: "submit",
+        variant: "pr",
+        content: "Add Color",
+      },
+    ],
+  },
+});
+
 const form = {
   fields: [
     {
@@ -234,7 +274,6 @@ export const toViewData = ({ state, props }, payload) => {
     ? toFlatItems(state.colorsData)
         .filter((item) => item.type === "color")
         .map((color) => ({
-          id: color.id,
           label: color.name,
           value: color.id,
         }))
@@ -245,7 +284,6 @@ export const toViewData = ({ state, props }, payload) => {
     ? toFlatItems(state.fontsData)
         .filter((item) => item.type === "font")
         .map((font) => ({
-          id: font.id,
           label: font.fontFamily,
           value: font.id,
         }))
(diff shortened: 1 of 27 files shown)


src/pages/typography/typography.store.js
```diff
@@ -199,16 +229,29 @@ export const toViewData = ({ state, props }, payload) => {
...
-    ? toFlatItems(state.colorsData)
-        .filter((item) => item.type === "color")
-        .map((color) => ({
-          id: color.id,
-          label: color.name,
-          value: color.id,
-        }))
-    : [];
+    ? [
+        ...toFlatItems(state.colorsData)
```
han4wluc: this is too nested.

do like this:

```
const colorOptions = ...
colorOptions.push({
  label: ...,
  value: ...
})
```

src/pages/typography/typography.store.js
```diff
@@ -199,16 +229,29 @@ export const toViewData = ({ state, props }, payload) => {
...
+          .filter((item) => item.type === "color")
+          .map((color) => ({
+            id: color.id,
```
han4wluc: we don't use id anymore. can remove

src/pages/typography/typography.store.js
```diff
@@ -221,6 +264,19 @@ export const toViewData = ({ state, props }, payload) => {
...
     : [];
 
+  // Generate folder options for add color dialog
+  const colorFolderOptions = [
+    { id: "_root", name: "Root Folder", value: "_root", label: "Root Folder" },
+    ...toFlatItems(state.colorsData)
+      .filter((item) => item.type === "folder")
+      .map((folder) => ({
+        id: folder.id,
+        name: folder.name || folder.id,
```
han4wluc: I think no need for `id` and `name`

src/pages/typography/typography.store.js
```diff
@@ -328,10 +384,50 @@ export const toViewData = ({ state, props }, payload) => {
         }
       : state.defaultValues;
 
+  // Add color dialog form
+  const addColorForm = {
```
han4wluc: having the entire config is too long and makes `toViewData` too hard to read.

define functino outside and do:

```
const addColorForm = createAddColoForm(colorFolderOptions)
```

han4wluc (commented):
the user interaction is correct.
but the 'add' feature is weird both in terms of UI and data structure.
we first need to change `rtgl-select` to support a proper 'add' feature from the component level.

738NGX:
@han4wluc please review the new implement

han4wluc:
there are some bugs:

* after color is created, it is being applied, but the select component is not showing it as selected.
* if you open the add color from again. it will display values from previous form

738NGX:
@han4wluc 

> after color is created, it is being applied, but the select component is not showing it as selected.

Now the color won't auto change. user need re-select what added

> if you open the add color from again. it will display values from previous form

fixed

src/pages/typography/typography.view.yaml
```diff
@@ -76,4 +76,5 @@ template:
           - rvn-font-preview flex=1 fontFamily=${previewFontFamily} previewText="${previewText}" fileId=${previewFontFileId} fontSize=${previewFontSize} lineHeight=${previewLineHeight} color=${previewColor} fontWeight=${previewFontWeight} width=350 showBorder=true:
 
   - rtgl-dialog#add-color-dialog ?open=${isAddColorDialogOpen} s=sm:
-    - rtgl-form#add-color-form slot=content .defaultValues=addColorDefaultValues .form=addColorForm w=f:
+    - $if isAddColorDialogOpen:
+      - rtgl-form#add-color-form slot=content .defaultValues=addColorDefaultValues .form=addColorForm w=f:
```
han4wluc, line 80: better updating key value rather than completely unmounting.

han4wluc:
new created color is not applied for the items preview. 
hypothesis is that the typography page does not have the newly added color in the state

https://github.com/user-attachments/assets/3157ef42-22e2-4729-afdb-22beb14d2c1b

src/pages/typography/typography.view.yaml
738NGX: Not work. key value update cannot update that

738NGX:
> new created color is not applied for the items preview. hypothesis is that the typography page does not have the newly added color in the state
> 
>  Screen.Recording.2025-08-20.at.18.25.46.mov

fixed now

han4wluc (approved):

PR #200, opened 2025-08-21
Character improve

738NGX:
- [x] Add character form, avatar should be below description
- [x] Fix cannot edit/upload a character
- [x] no need to display those: Type, File Type, File Size
- [ ] Can double click item to pop up form and edit
  - This have the same function as right detail pannel. Maybe double click to enter the sprite button instead of click the "sprite" button is better, because other page like layout group view do the same things.

And now it's empty in a new character's sprite page and has no hint. Need add something?


han4wluc:
> And now it's empty in a new character's sprite page and has no hint. Need add something?

yes, when character is created. we should create it with an empty folder.

han4wluc (approved):

PR #202, opened 2025-08-22
Feat duplicate

738NGX:


src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -167,6 +167,113 @@ export const handleFileAction = (e, deps) => {
         },
       },
     });
+  } else if (item.value === "duplicate-item") {
+    const repositoryState = repository.getState();
+    const targetData = lodashGet(repositoryState, repositoryTarget);
+    const currentItem =
+      targetData && targetData.items ? targetData.items[itemId] : null;
+
+    if (currentItem) {
```
han4wluc: negative condition first
```
if (!currentItem) {
  return;
}
```

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -167,6 +167,113 @@ export const handleFileAction = (e, deps) => {
...
+        }
+        return null;
+      };
+
+      const parentId = targetData.tree
+        ? findParentId(targetData.tree, itemId)
+        : "_root";
+
+      // Helper function to duplicate an item and its children recursively
+      const duplicateItemRecursive = (itemToDuplicate, isRoot = true) => {
```
han4wluc: this algorithm is a bit long.
have you considered other simpler solution?

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -167,6 +163,86 @@ export const handleFileAction = (e, deps) => {
...
+
+    // Create duplicated item with new ID and name
+    const duplicatedItem = {
+      ...currentItem,
+      id: nanoid(),
+      name: `${currentItem.name} (copy)`,
+    };
+
+    // Add the duplicated item
+    repository.addAction({
```
han4wluc, line 177: I don't like that we're creating a lot of 'treePush' actions.

what about we implement duplicate natively inside the repository?

```
repository.addAction({
  actionType: 'copy',
  target: ...,
  value: {
    // some way to specify the position. default to immediately after the current item.
  }
}
```

738NGX:
The key reason is the random id will not always same when do `treeCopy`:

// Initial execution

treeCopy action -> Generates ID: "uW0hFNMt2o7SWLneS3TIE"
Saves to localStorage

// Page refresh, replays actions

treeCopy action -> Generates new ID: "WhtGGPcjc_hRSEUNZO9kz" (different!)

// User attempts to delete

The UI sends a delete request using ID: "uW0hFNMt2o7SWLneS3TIE" (old)

But the state only contains: "WhtGGPcjc_hRSEUNZO9kz" (new)

Delete failed

So the solution is either add new id in local storage or extend the origin id

now the new id is the origin id+'_copy', or _copy1, _copy2, ... This will keep the id always same

han4wluc:
> The key reason is the random id will not always same when do `treeCopy`:
> 
> // Initial execution
> 
> treeCopy action -> Generates ID: "uW0hFNMt2o7SWLneS3TIE" Saves to localStorage
> 
> // Page refresh, replays actions
> 
> treeCopy action -> Generates new ID: "WhtGGPcjc_hRSEUNZO9kz" (different!)
> 
> // User attempts to delete
> 
> The UI sends a delete request using ID: "uW0hFNMt2o7SWLneS3TIE" (old)
> 
> But the state only contains: "WhtGGPcjc_hRSEUNZO9kz" (new)
> 
> Delete failed
> 
> So the solution is either add new id in local storage or extend the origin id
> 
> now the new id is the origin id+'_copy', or _copy1, _copy2, ... This will keep the id always same

I see the problem now. makes sense.

`_copy', or _copy1, _copy2`

what is best solution based on those requirements:
* be careful if you copy the item twice, it should create duplicated ids
* I would like them to be still random ids

738NGX:
@han4wluc 

> I would like them to be still random ids

Now when copy items, the new random id will auto generate and save it in local storage. When next execute after refresh, can detect generated ids, won't generate again.

src/deps/repository.js
```diff
@@ -198,6 +200,90 @@ const treeDelete = (state, target, value) => {
...
+    return newState; // Node not found, return unchanged state
+  }
+
+  const { node, parent } = nodeInfo;
+
+  // Check if this action already has generated IDs (from a previous execution)
+  // If so, use them; otherwise generate new ones and store them
+  // This allows us to use random IDs but keep them consistent across replays
+  if (!value._generatedIds) {
+    value._generatedIds = {};
```
han4wluc: it is not allowed to mutate function argument. this is causing a side effect, and will cause all sorts of problems.

han4wluc: instead of giving 1 solution each time. could you give all possible solutions you can think of and then we pick the best one?

738NGX: It's becoming a paradox, if must elimate side effect and use random id, can only do that by adding more things in value instead of just an id.

Solution 1 is generate ids in fileExplorer handler and pass them in value

Solution 2 is use non-random id

han4wluc: > Solution 1 is generate ids in fileExplorer handler and pass them in value.

lets go with this one

but only the id comes from outside, the copying logic I want to keep inside the repository.


technically there is another solution where we pass a seed value. then the random generation becomes deterministic. but im not sure if nanoid supporta that.

738NGX: nanoid doesn't have built-in support for seeded/deterministic ID generation. 

But it provides `customRandom` api which can help implement that, for example:

```js
const createSeededIdGenerator = (seed) => {
  let currentSeed = seed;
  
  const random = (size) => {
    const bytes = new Uint8Array(size);
    for (let i = 0; i < size; i++) {
      // Linear congruential generator
      currentSeed = (currentSeed * 1103515245 + 12345) & 0x7fffffff;
      bytes[i] = currentSeed & 0xff;
    }
    return bytes;
  };
  
  const alphabet = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_-';
  return customRandom(alphabet, 21, random);
};
```

if that is ok, can simply pass the timestamp as seed in value and generate ids

han4wluc: yes. can try that

738NGX:
implemented

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -167,6 +163,25 @@ export const handleFileAction = (e, deps) => {
...
+      return;
+    }
+
+    // Add the duplicated item with timestamp as seed
+    repository.addAction({
+      actionType: "treeCopy",
+      target: repositoryTarget,
+      value: {
+        id: itemId,
+        seed: Date.now(),
```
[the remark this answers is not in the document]
738NGX: the seed can only be number, so the nanoid must be hashed if want to use

```
if (typeof seed === 'string') {
    // Simple hash function for string to number conversion
    currentSeed = 0;
    for (let i = 0; i < seed.length; i++) {
      const char = seed.charCodeAt(i);
      currentSeed = ((currentSeed << 5) - currentSeed) + char;
      currentSeed = currentSeed & 0x7fffffff; // Keep it positive 31-bit
    }
  }
```

is that ok?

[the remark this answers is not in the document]
738NGX: now is

```
value: {
  id: itemId,
  seed: Math.floor(Math.random() * 0x7fffffff),
},
```

PR #216, opened 2025-08-28
Tauri entry

738NGX:
Environment variables are not available in the browser. The alternative here is to run a script at build time to automatically generate setup.js instead of dynamically importing based on environment variables.


han4wluc (commented):
I checked your PR and current code. solution is this:


* `rettangoli-fe` can take path to `setup.js` 
https://github.com/yuusoft-org/rettangoli/blob/23cf53dc400e151e69b75a22caa77913d2244e21/packages/rettangoli-fe/src/cli/build.js#L50

* update `rettangoli-cli` to be able to specify a custom path from `setup.js`. which takes priority over `config.fe.setup`
https://github.com/yuusoft-org/rettangoli/blob/23cf53dc400e151e69b75a22caa77913d2244e21/packages/rettangoli-cli/cli.js#L57

* to test it locally, I usually do something like this:

`bun run ../rettangoli/packages/rettangoli-cli/cli.js fe`

This way we, can do 

* `rtgl fe build` (will use default `setup.js` or one specified in `rettangoli.config.yaml`)
* `rtgl fe build -s src/setup.tauri.js` or `rtgl fe build --setup-path src/setup.tauri.js`
* `rtgl fe build -s src/setup.web.js` or `rtgl fe build --setup-path src/setup.web.js`

---

remove `setup.common.js` and `build-setup.js`, we only use `setup.web.js` and `setup.tauri.js`.

`setup.web.js` and `setup.tauri.js` will have some duplicate code.
* stuff that is just import and export it is ok to be duplicate.
* but function implmenetations like `fetchTemplateFonts` and `fetchTemplateImages` should be moved out of setup.js. setup.js should only import the function.

README.md
```diff
@@ -155,5 +174,21 @@ The built application will be available in `src-tauri/target/release/` with plat
...
+
+### Platform-Specific Entry Points
+
+The project implements separate entry points for web and desktop builds:
+
+- **`src/setup.common.js`** - Contains all shared initialization logic
+- **`src/setup.web.js`** - Web-specific configuration (uses IndexedDB, web APIs)
+- **`src/setup.tauri.js`** - Desktop-specific configuration (can access Tauri APIs, native filesystem)
+
+During build, the appropriate platform file is automatically selected using the `scripts/build-setup.js` script. This allows for:
```
han4wluc: this no more needed

han4wluc (approved):

PR #217, opened 2025-08-29
Tauri feat sqlite

738NGX:
Notes:
- `template.js` was created incorrectly in the previous PR and caused some problems, now it had been removed. The template load logic back to the setup.xx.js
- Refactored `createRepository`
- Now Web Actions saved in IndexedDB, Tauri Actions saved in local sqlite
- File system has not been implemented in this PR, Tauri still use IndexedDB to save files now because there is no conflict with sqlite.


src/deps/repository.js
```diff
@@ -459,6 +445,14 @@ const createRepositoryInternal = (
     actionStream.push(action);
     // Invalidate cache when new action is added
     isCacheValid = false;
+
+    // Persist to storage if adapter is provided (fire-and-forget)
+    if (storageAdapter && storageAdapter.addAction) {
```
han4wluc: no need to check if storageAdapter and storageAdapter.addAction if we already assert it exists earlier during creation.

src/deps/repository.js
```diff
@@ -459,6 +445,14 @@ const createRepositoryInternal = (
...
+      // Don't await - let it save in the background
+      storageAdapter.addAction(action).catch((error) => {
```
han4wluc: it should be async. we let the user decide whether to catch or not for `repository.addAction`

src/deps/repository.js
```diff
@@ -513,13 +507,27 @@ const createRepositoryInternal = (
...
+  const flush = async () => {
+    // Save all actions to storage adapter
+    if (storageAdapter && storageAdapter.saveAllActions) {
+      await storageAdapter.saveAllActions(actionStream);
+    }
+  };
+
+  const init = async () => {
+    // Load all events from storage adapter
+    if (storageAdapter && storageAdapter.getAllEvents) {
```
han4wluc: dont do this. it will fail silently.

src/deps/repository.js
```diff
@@ -513,13 +507,27 @@ const createRepositoryInternal = (
     return actionStream;
   };
 
-  const flush = () => {
-    if (localStorageKey) {
-      localStorage.setItem(localStorageKey, JSON.stringify(actionStream));
+  const flush = async () => {
+    // Save all actions to storage adapter
+    if (storageAdapter && storageAdapter.saveAllActions) {
+      await storageAdapter.saveAllActions(actionStream);
```
han4wluc: flush is not needed anymore if we already persist during `addAction`

src/setup.tauri.js
```diff
@@ -124,47 +124,202 @@ const initialData = {
...
+      return results.map((row) => ({
+        actionType: row.action_type,
+        target: row.target,
+        value: row.value ? JSON.parse(row.value) : null,
+      }));
+    },
+
+    async saveAllActions(actions) {
+      // Clear existing actions and insert new ones in a transaction
+      await db.execute("DELETE FROM actions");
```
han4wluc: no... never do this. if it crashes after this before inserts. we have wiped all user's data...

han4wluc: lets remove `saveAllActions` alltogether. no needed maymore.

src/setup.tauri.js
```diff
@@ -124,47 +124,202 @@ const initialData = {
...
+  imageTree.push({
+    id: folderId,
+    children: folderChildren,
   });
 
+  return { fetchedImages, imageItems, imageTree };
+}
+
+// Fetch template fonts from static folder
+async function fetchTemplateFonts() {
```
han4wluc: we would still want to have same function for both tauri and web. what is the difference?

src/setup.tauri.js
```diff
@@ -124,47 +124,202 @@ const initialData = {
   },
 };
 
-const localStorageKey = "repositoryEventStream";
-const repository = createRepository(initialData, localStorageKey);
+// Create Tauri SQLite Repository Adapter
+const createTauriSQLiteRepositoryAdapter = async () => {
```
han4wluc: move to separate file under `deps` folder

src/setup.web.js
```diff
@@ -112,47 +111,226 @@ const initialData = {
   },
 };
 
-const localStorageKey = "repositoryEventStream";
-const repository = createRepository(initialData, localStorageKey);
+const createIndexeddbRepositoryAdapter = () => {
```
han4wluc: move to separate file under `deps` folder

src/deps/repository.js
```diff
@@ -459,6 +445,14 @@ const createRepositoryInternal = (
     actionStream.push(action);
```
han4wluc, line 449: do this after db succeeded. in case db failed.

src/deps/repository.js
738NGX: if await here, when call `render` the database hasn't been written yet, so either change all handlers to async, or keep updating memory syncly and writing to database asyncly

src/deps/repositoryAdapters.js
```diff
@@ -0,0 +1,99 @@
...
+    },
+  };
+};
+
+/**
+ * Tauri SQLite Repository Adapter
+ */
+export const createTauriSQLiteRepositoryAdapter = async () => {
+  // Dynamic import to avoid loading in non-Tauri environments
+  const Database = (await import("@tauri-apps/plugin-sql")).default;
```
han4wluc: dont do dynamic import.

the tauri implementation should be in a separate file that is imported only in the tauri build

src/deps/repository.js
han4wluc: hmmm yes. for now make it async. but we don't call async in the handlers. maybe will chenge it later.

han4wluc (approved):
all good except the dynamic import thing.

PR #218, opened 2025-09-01
Tauri feat fs

738NGX:
Note:
- have a change on `fileExplorer.handlers.js`, because in indexedDB we don't remove the deleted files. Now in fs implemented remove. If it's unnecessary, can also remove that logic.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,6 +4,7 @@
     "": {
       "dependencies": {
         "@rettangoli/fe": "^0.0.7-rc11",
+        "@tauri-apps/api": "^2.8.0",
         "@tauri-apps/plugin-fs": "^2.4.2",
         "@tauri-apps/plugin-sql": "~2",
         "esbuild": "^0.25.5",
--- a/bun.lock
+++ b/bun.lock
@@ -4,7 +4,6 @@
     "": {
       "dependencies": {
         "@rettangoli/fe": "^0.0.7-rc11",
-        "@tauri-apps/api": "^2.8.0",
         "@tauri-apps/plugin-fs": "^2.4.2",
         "@tauri-apps/plugin-sql": "~2",
         "esbuild": "^0.25.5",
--- a/bun.lock
+++ b/bun.lock
@@ -4,6 +4,8 @@
     "": {
       "dependencies": {
         "@rettangoli/fe": "^0.0.7-rc11",
+        "@tauri-apps/api": "^2.8.0",
+        "@tauri-apps/plugin-fs": "^2.4.2",
         "@tauri-apps/plugin-sql": "~2",
         "esbuild": "^0.25.5",
         "immer": "^10.1.1",
@@ -157,6 +159,8 @@
     "@tauri-apps/cli-win32-ia32-msvc": ["@tauri-apps/cli-win32-ia32-msvc@2.8.3", "", { "os": "win32", "cpu": "ia32" }, "sha512-aIP38i2KeADboPD1wsBFYdodEQ9PIJe0HW2urd3ocHpGxF8gX/KMiGOwGVSobu9gFlCpFNoVwCX6J1S5pJCRIQ=="],
 
     "@tauri-apps/cli-win32-x64-msvc": ["@tauri-apps/cli-win32-x64-msvc@2.8.3", "", { "os": "win32", "cpu": "x64" }, "sha512-Z+H+PwK+3yMffG1rN7iqs+uPo6FkPyHJ4MTtFhnEvvGzc3aH711bwFb6+PXwMXfOb/jPR/LB+o6kEXghBu9ynQ=="],
+
+    "@tauri-apps/plugin-fs": ["@tauri-apps/plugin-fs@2.4.2", "", { "dependencies": { "@tauri-apps/api": "^2.8.0" } }, "sha512-YGhmYuTgXGsi6AjoV+5mh2NvicgWBfVJHHheuck6oHD+HC9bVWPaHvCP0/Aw4pHDejwrvT8hE3+zZAaWf+hrig=="],
 
     "@tauri-apps/plugin-sql": ["@tauri-apps/plugin-sql@2.3.0", "", { "dependencies": { "@tauri-apps/api": "^2.6.0" } }, "sha512-JYwIocfsLaDWa41LMiZWuzts7yCJR+EpZPRmgpO7Gd7XiAS9S67dKz306j/k/d9XntB0YopMRBol2OIWMschuA=="],
 
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -22,9 +22,13 @@
       }
     ],
     "security": {
-      "csp": null,
-      "dangerousDisableAssetCspModification": true,
-      "devCsp": null
+      "csp": "default-src 'self' ipc: http://ipc.localhost; img-src 'self' asset: http://asset.localhost data: blob:; media-src 'self' asset: http://asset.localhost data: blob:; font-src 'self' asset: http://asset.localhost data:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'",
+      "dangerousDisableAssetCspModification": false,
+      "devCsp": null,
+      "assetProtocol": {
+        "enable": true,
+        "scope": ["$APPDATA/**", "**"]
+      }
     }
   },
   "bundle": {
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -22,8 +22,8 @@
       }
     ],
     "security": {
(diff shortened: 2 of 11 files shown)


src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,312 @@
...
+  exists,
+  mkdir,
+  remove,
+  BaseDirectory,
+} from "@tauri-apps/plugin-fs";
+import { detectFileType } from "./fileProcessors";
+
+// Generate unique file ID
+const generateFileId = () => {
+  if (typeof crypto !== "undefined" && crypto.randomUUID) {
```
738NGX: Here copied the logic in indexedDb, maybe can both change into nanoid

han4wluc: yes, can change to nanoid

bun.lock
```diff
@@ -4,6 +4,8 @@
     "": {
       "dependencies": {
         "@rettangoli/fe": "^0.0.7-rc11",
+        "@tauri-apps/api": "^2.8.0",
```
han4wluc, line 7: do we use this? what is this for?

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -121,6 +121,22 @@ export const handleFileAction = (e, deps) => {
       targetData && targetData.items ? targetData.items[itemId] : null;
 
     if (currentItem) {
+      // If the item has a fileId, delete the actual file from storage
+      if (currentItem.fileId && deps.fileManager) {
+        //console.log(`[FileExplorer] Deleting file with fileId: ${currentItem.fileId}`);
+        try {
+          // Use fileManager's deleteFile method which calls storageAdapter.deleteFile
+          await deps.fileManager.deleteFile(currentItem.fileId);
```
han4wluc: no need to delete.
if we don't delete, we can enable versioning, and time travel to check old versions.
similar to git. if we delete it here, the file is gone forever and user cannot restore.

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,312 @@
...
+// Get file extension from filename
+const getExtension = (filename) => {
+  if (!filename) return "bin";
+  const parts = filename.split(".");
+  return parts.length > 1 ? parts.pop().toLowerCase() : "bin";
+};
+
+// Get subdirectory based on file type
+// Maps detectFileType results to directory names
+const getFileSubdir = (file) => {
```
han4wluc: no subdirectories.
store it as `${folder}/${fileId}`, if possible I would even skip the file extension.

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,312 @@
...
+          recursive: true,
+        });
+      }
+    } catch (error) {
+      console.error("Failed to create directories:", error);
+    }
+  };
+
+  // Read manifest file
+  const readManifest = async () => {
```
han4wluc: do we need manifest file? is it possible to implement without manifest file?

I think it should be possible, current web version does not use any centralized manifest .

I would like to remove all manifest related logic

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,312 @@
...
+  return typeToDir[fileType] || "misc";
+};
+
+// Create the Tauri File System storage adapter
+export const createTauriFileSystemStorageAdapter = () => {
+  const urlCache = new Map();
+
+  // For now use hard coded path, in the future this func can receive parameter like `projectManager` to handle more paths
+  const PROJECT_PATH = "routevn-creator/projects/default";
+  const ASSETS_PATH = `${PROJECT_PATH}/assets`;
```
han4wluc: const FILES_PATH = `${PROJECT_PATH}/files`;

738NGX:
All fixed

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+
+      // Log for debugging
+      console.log(
+        `[TauriFS] File ${fileId} marked as deleted but preserved on disk for versioning`,
+      );
+    },
+
+    // List all files for a project
+    async listFiles(projectId) {
+      // Since we don't have a manifest, we can't really list files
```
han4wluc: remove all those comments

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -121,6 +121,9 @@ export const handleFileAction = (e, deps) => {
       targetData && targetData.items ? targetData.items[itemId] : null;
 
     if (currentItem) {
+      // Don't delete the actual file - keep it for versioning/time travel
```
han4wluc: remove all those comments

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+  remove,
+  BaseDirectory,
+} from "@tauri-apps/plugin-fs";
+import { nanoid } from "nanoid";
+
+// Create the Tauri File System storage adapter
+export const createTauriFileSystemStorageAdapter = () => {
+  const urlCache = new Map();
+
+  // For now use hard coded path, in the future this func can receive parameter like `projectManager` to handle more paths
```
han4wluc: remove all those comments

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  return {
+    // Store a file in the file system
+    async storeFile(file, projectId) {
+      try {
+        await ensureDirectories();
+
+        const fileId = nanoid();
+        // Store without extension for simplicity
```
han4wluc: remove all those comments

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+        const arrayBuffer = await file.arrayBuffer();
+        const uint8Array = new Uint8Array(arrayBuffer);
+
+        // Write file to disk - flat structure
+        const filePath = `${FILES_PATH}/${fileName}`;
+        await writeFile(filePath, uint8Array, {
+          baseDir: BaseDirectory.AppData,
+        });
+
+        // Generate access URL - use blob URL for now to match indexedDBStorageAdapter
```
han4wluc: remove all those comments

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+
+        // Reuse storeFile for metadata
+        return await this.storeFile(jsonBlob, projectId);
+      } catch (error) {
+        console.error("Failed to store metadata:", error);
+        throw error;
+      }
+    },
+
+    // Delete file from file system - NO LONGER DELETES FILES
```
han4wluc: remove all those comments

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+    // Files are kept for versioning/time travel (like git)
+    async deleteFile(fileId) {
+      // Don't actually delete the file
```
han4wluc: remove all those comments

han4wluc (commented):
code is ok. if the comment is obvious or just repeating the code, delete the comment.
remember code is to describe things that are non obvious from code.

han4wluc (approved):

src/deps/tauriFileSystemStorageAdapter.js
```diff
@@ -0,0 +1,168 @@
...
+        const uint8Array = new Uint8Array(arrayBuffer);
+
+        // Write file to disk - flat structure
+        const filePath = `${FILES_PATH}/${fileName}`;
+        await writeFile(filePath, uint8Array, {
+          baseDir: BaseDirectory.AppData,
+        });
+
+        // Generate access URL - use blob URL for now to match indexedDBStorageAdapter
+        const downloadUrl = URL.createObjectURL(file);
```
han4wluc: note that this will load data into RAM memory.
currently you are adding everything to memory. if we have 100MB of files, it in theory will consume 100MB RAM, if you have 1 GB of files it will load 1 GB into memory.
this will happen when you open the images page.

so... we cannot accept this. either remove the cache alltogether or make sure we limit the amount of memory to cache but still be useful.

738NGX: Now no longer use cache, use `@tauri-apps/api` serve an asset url to access files

https://v2.tauri.app/zh-cn/reference/javascript/api/namespacecore/#convertfilesrc

src-tauri/tauri.conf.json
```diff
@@ -22,9 +22,13 @@
       }
     ],
     "security": {
-      "csp": null,
-      "dangerousDisableAssetCspModification": true,
-      "devCsp": null
+      "csp": "default-src 'self' ipc: http://ipc.localhost; img-src 'self' asset: http://asset.localhost data: blob:; media-src 'self' asset: http://asset.localhost data: blob:; font-src 'self' asset: http://asset.localhost data:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'",
```
han4wluc: this seems too many. can you only add the ones that we need?
in security perspective, the more things you add, the unsafer it becomes.

han4wluc (approved):

PR #220, opened 2025-09-02
Feat project manager

738NGX:


PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+        │   └── files/           # Project-specific file storage (flat structure)
+        └── {projectId}/
+            └── ...
+```
+
+### Database Schema
+
+#### Projects Management Database (projects.db)
+```sql
+CREATE TABLE projects (
```
han4wluc: use sqlite, but use only key value table.

```
CREATE TABLE ItemTable (
    key TEXT PRIMARY KEY,
    value BLOB or TEXT
)
```

reason, is that when we need to store new things, we don't want to do a db migration.


key: will be something like `projects`
value: the json with projects data.

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
+# Project Creation Feature Implementation Plan
+
+## Architecture Design
+
+### Data Storage Structure
+```
+AppData/
+└── routevn-creator/
+    ├── projects.db              # Project management database
+    └── projects/
```
han4wluc: projects folder needs to be outside AppData

AppData is usually for the application data. not user data.

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+};
+
+export const setProjects = (state, projects) => {
+    state.projects = projects;
+};
+```
+
+### Phase 2: User Interface Components
+
+#### 2.1 Project Creation Dialog
```
han4wluc: overkill. just use existing `pages/projects`

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+    const templates = {
+        'default': { /* existing template data */ },
+        'template1': { /* template 1 */ },
+        'template2': { /* template 2 */ }
+    };
+    return templates[templateId] || templates['default'];
+};
+```
+
+### Phase 4: Application Lifecycle
```
han4wluc: we can skip all this for now. lets just focus 1 step at a time, lets build projects page first, and then work on those next stuff

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
han4wluc: you will need to create a new file under deps for this key value set and get

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Implementation Steps
+
+### Phase 1: Project Management Infrastructure
+
+#### 1.1 Create Project Management Service
+**File:** `src/services/projectManager.js`
+
+```javascript
+export const createProjectManager = (deps) => {
```
han4wluc: no needed. just use a key value store. the kv will store to the app sqlite.

```js
const projects = await kv.get('projects')
projects.push(project)
await kv.set('projects', projects)
```

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+        createProject,
+        openProject,
+        listProjects,
+        deleteProject,
+        getLastOpenedProject
+    };
+};
+```
+
+#### 1.2 Update Repository Adapter
```
han4wluc: lets think about this later.

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+**File:** `src/deps/tauriRepositoryAdapter.js`
+
+Modifications:
+- Accept projectId in constructor
+- Use project-specific database path
+- Update file storage paths to include projectId
+
+#### 1.3 Extend App Store for Project Management
```
han4wluc: lets think about this later

PROJECT_CREATION_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,200 @@
...
+**Files:** Update existing `src/pages/projects/`
+
+Features:
+- List view of all projects
+- Project cards showing: name, icon, description, last opened date
+- Quick actions: Open, Delete
+- "Create New Project" button
+
+
+### Phase 3: Template System Extension
```
han4wluc: lets think about this later.

src/pages/projects/projects.handlers.js
```diff
@@ -1,8 +1,14 @@
...
-//   deps.httpClient.get('...').then(response => {})
-//   deps.setSubscriptions([])
-// }
+import { nanoid } from "nanoid";
+
+export const handleBeforeMount = (deps) => {
+  const { keyValueStore, store, render } = deps;
+
+  // Load projects from key-value store (async but don't await)
+  keyValueStore.get("projects").then((projects) => {
```
han4wluc: if is async, put it in `handleAfterMount`

src/pages/projects/projects.handlers.js
```diff
@@ -27,9 +33,113 @@ export const handleCloseDialogue = (payload, deps) => {
...
     path: `/project`,
   });
 };
+
+export const handleBrowseFolder = async (e, deps) => {
+  const { store, render } = deps;
+
+  try {
+    // Import Tauri dialog API
+    const { open } = await import("@tauri-apps/plugin-dialog");
```
han4wluc: create a file in deps

src/pages/projects/projects.handlers.js
```diff
@@ -27,9 +33,113 @@ export const handleCloseDialogue = (payload, deps) => {
...
+      title: "Select Project Location",
+    });
+
+    if (selected) {
+      // Update the form's default value for projectPath
+      store.setProjectPath(selected);
+      render();
+    }
+  } catch (error) {
+    console.error("Error selecting folder:", error);
```
han4wluc, line 63: use `alert` so is not silent failing

src/pages/projects/projects.handlers.js
```diff
@@ -27,9 +33,113 @@ export const handleCloseDialogue = (payload, deps) => {
...
+    // Slot fields need to be retrieved from store using select function
+    const projectPath = store.selectProjectPath();
+
+    // Validate input
+    if (!name || !description || !projectPath) {
+      return;
+    }
+
+    // Generate a simple ID
+    const projectId = "project_" + nanoid();
```
han4wluc: for desktop, there is no projectId.
use something like deviceProjectId or tempProjectId. we only store this in the app sqlite. but is not for backend or anything.

when user remove a project, and open the path again to create a new project. the id should be a new one.

han4wluc (approved):

PR #221, opened 2025-09-03
feat project manager global database

738NGX:
The rust code is mainly a connection to `rusqlite` library, different to tauri provided api, its a rust origin lib. It provides command, and use invoke api provided by tauri connect to js code. Nothing is implementing from scratch.

The current status is can correctly create .db file at select folder and write `init` action into .db file. the template file copy and load project haven't implemented yet


src-tauri/src/project_db.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        let db_path = PathBuf::from(project_path).join("repository.db");
+        
+        std::fs::create_dir_all(project_path).ok();
+        
+        let conn = Connection::open(db_path)?;
+        
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS actions (
```
han4wluc: this is application data. this should belong to js not rust.

src-tauri/src/project_db.rs
```diff
@@ -0,0 +1,138 @@
...
+        Ok(())
+    }
+
+    pub fn add_action(&self, project_path: &str, action: Action) -> Result<()> {
+        let conns = self.connections.lock().unwrap();
+        let conn = conns.get(project_path)
+            .ok_or(rusqlite::Error::InvalidPath(PathBuf::from(project_path)))?;
+        
+        conn.execute(
+            "INSERT INTO actions (action_type, target, value) VALUES (?1, ?2, ?3)",
```
han4wluc: keep application logic in js

src/deps/tauriRepositoryAdapter.js
```diff
@@ -1,40 +1,42 @@
-import Database from "@tauri-apps/plugin-sql";
+import { invoke } from "@tauri-apps/api/core";
 
 /**
  * Tauri SQLite Repository Adapter
+ * @param {string} projectPath - Required project path for project-specific database
  */
-export const createTauriSQLiteRepositoryAdapter = async () => {
-  // Initialize SQLite database
-  const db = await Database.load("sqlite:repository.db");
```
han4wluc, line 8: keep this exact interface. should be very easy to replace one sqlite client from other.

src/pages/projects/projects.handlers.js
```diff
@@ -93,15 +104,108 @@ export const handleFormSubmit = async (e, deps) => {
...
+          typography: { items: {}, tree: [] },
+          variables: { items: {}, tree: [] },
+          components: { items: {}, tree: [] },
+          layouts: { items: {}, tree: [] },
+          preset: { items: {}, tree: [] },
+          scenes: { items: {}, tree: [] },
+        };
+
+        // Initialize repository to add template data
+        const { createTauriSQLiteRepositoryAdapter } = await import(
```
han4wluc: i think I told you many times already, we do not want dynamic imports. please never use dynamic imports unless there is a good reason to do it.

src/pages/projects/projects.handlers.js
```diff
@@ -39,10 +47,13 @@ export const handleProjectsClick = async (e, deps) => {
...
   await keyValueStore.set("lastOpenedProjectId", id);
 
-  // Navigate to project page
+  // Navigate to project page - the page will need to be refreshed to use the new database
   subject.dispatch("redirect", {
     path: `/project`,
   });
+
+  // Reload the page to reinitialize with the project database
+  window.location.reload();
```
han4wluc: no. we don't want to reload the page.

src/pages/projects/projects.handlers.js
```diff
@@ -1,4 +1,12 @@
 import { nanoid } from "nanoid";
+import { invoke } from "@tauri-apps/api/core";
```
han4wluc: I don't want all those side effects in the handler.
side effects needs to be in deps.

src/pages/projects/projects.handlers.js
```diff
@@ -93,15 +104,108 @@ export const handleFormSubmit = async (e, deps) => {
...
+        };
+
+        // Initialize repository to add template data
+        const { createTauriSQLiteRepositoryAdapter } = await import(
+          "../../deps/tauriRepositoryAdapter"
+        );
+        const { createRepository } = await import("../../deps/repository");
+
+        const adapter = await createTauriSQLiteRepositoryAdapter(projectPath);
+        const repo = createRepository(initialData, adapter);
```
han4wluc: this is too weird and will not work.
user can open for example images page direclty without going through projects page.

src/setup.tauri.js
```diff
@@ -116,14 +114,13 @@ const initialData = {
...
-// Initialize adapter and repository
-// TODO: This should be moved to project opening logic
-// Currently creates repository.db in AppData, should be in project folder
-const repositoryAdapter = await createTauriSQLiteRepositoryAdapter();
-const repository = createRepository(initialData, repositoryAdapter);
-
-// Initialize repository with stored data
-await repository.init();
+// Create a placeholder repository - will be initialized when entering a project page
+const repository = {
```
han4wluc: we still want to pass repository globally. but you might need to wrap it with project.

for example
```
repository.addAction(projectId, ...)
```

738NGX:
Have refactored

src/deps/repository.js
```diff
@@ -420,49 +421,73 @@ const treeMove = (state, target, value) => {
   return newState;
 };
 
-export const createRepository = (initialState, storageAdapter) => {
-  // storageAdapter is required for persistence
-  if (!storageAdapter) {
-    throw new Error("Storage adapter is required");
-  }
+// Store project paths
+const projectPaths = new Map();
```
han4wluc: you are not allowed to store global var here. this is side effect makes whole pure functions meaningless

src-tauri/src/project_db.rs
```diff
@@ -0,0 +1,122 @@
+use rusqlite::{Connection, Result};
```
han4wluc: this custom implementation is not good enough. the API is too different from original one. too much custom code to maintain and prone to issues.

lets just keep use original sqlite library. and save all projects under AppData for now.

meanwhile we will try to get Tauri to officially support absolute paths. I've commented in the issue: https://github.com/tauri-apps/plugins-workspace/issues/275#issuecomment-3249321270

han4wluc (commented):
this is getting messy, especially all the dependencies. lets look at it together tomorrow.

src-tauri/Cargo.toml
```diff
@@ -27,3 +27,5 @@ tauri-plugin-devtools = "2"
 tauri-plugin-sql = { version = "2", features = ["sqlite"] }
```
han4wluc: i've forked the plugin source code:

try this:

```
tauri-plugin-sql = { git = "https://github.com/han4wluc/plugins-workspace", branch = "sql-absolute-path" }
```

then:
```
await Database.load("sqlite:/path_to_aboslute_path/repository.db");
```

code: https://github.com/tauri-apps/plugins-workspace/compare/v2...han4wluc:plugins-workspace:sql-absolute-path?expand=1

738NGX: this can work

src/deps/repository.js
han4wluc: can you fix this one as well?

738NGX: will add `projectPath` parameter directly to `createRepository` function, then the global var will no longer use. can this work?

han4wluc: if you do that. will you need to call `createRepository` everytime you call `addAction`? that would not be good.

han4wluc (approved):

738NGX:
Now web code directly use `export const createRepository = createRepositoryInternal;` can work with the refactored api

738NGX:
Refactor Completed

han4wluc (approved):
ok as long as it works during manual testing

PR #223, opened 2025-09-05
Bug fixes

738NGX:
Fixes:

- The template use file id as imageId, this is a mistake leads to sprite image doesnt render
- Font Page siderbar has a different width from others
- scene didnt show correctly, p didnt pass in payload at scene editor

Still problems:

scene editor white screen, though the data had been loaded

image


han4wluc (approved):
any error logs in the scene editor ?

738NGX:
> any error logs in the scene editor ?

I knew, there's a async logic in before mount, now fixed

PR #226, opened 2025-09-09
Audio improve

738NGX:
- fix waveform cannot display under refactored system
- improve waveform file format to reduce size

### DIFF
--- a/src/components/waveformVisualizer/waveformVisualizer.handlers.js
+++ b/src/components/waveformVisualizer/waveformVisualizer.handlers.js
@@ -1,15 +1,22 @@
 export const handleAfterMount = async (deps) => {
-  const { attrs, store, render, downloadWaveformData } = deps;
+  const { attrs, store, render, fileManagerFactory, router } = deps;
 
   if (!attrs.waveformDataFileId) {
     return;
   }
 
   try {
-    // Download waveform data from API Object Storage
-    const waveformData = await downloadWaveformData({
+    const { p: projectId } = router.getPayload();
+    const fileManager = await fileManagerFactory.getByProject(projectId);
+
+    const waveformData = await fileManager.downloadMetadata({
       fileId: attrs.waveformDataFileId,
     });
+
+    // Convert byte data (0-255) back to normalized values (0-1)
+    if (waveformData && waveformData.data) {
+      waveformData.data = waveformData.data.map((value) => value / 255);
+    }
 
     store.setWaveformData(waveformData);
     store.setLoading(false);
@@ -22,7 +29,7 @@ export const handleAfterMount = async (deps) => {
 };
 
 export const handleOnUpdate = async (changes, deps) => {
-  const { attrs, store, render, downloadWaveformData } = deps;
+  const { attrs, store, render, fileManagerFactory, router } = deps;
 
   if (!attrs.waveformDataFileId) {
     return;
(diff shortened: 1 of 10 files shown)


src/components/waveformVisualizer/waveformVisualizer.handlers.js
```diff
@@ -1,16 +1,23 @@
...
+    const { p: projectId } = router.getPayload();
+    const fileManager = await fileManagerFactory.getByProject(projectId);
+
+    const waveformData = await fileManager.downloadMetadata({
       fileId: attrs.waveformDataFileId,
     });
 
+    // Convert byte data (0-255) back to normalized values (0-1)
+    if (waveformData && waveformData.data) {
+      waveformData.data = waveformData.data.map((value) => value / 255);
```
han4wluc: no transformation needed when we download.
we should transform when we upload & write it. 
this way there is no inconsistency issue, currently you're doing this transfomration in 3 places.
also we want to reduce filesize.

src/deps/indexedDBStorageAdapter.js
```diff
@@ -205,9 +206,12 @@ export const createIndexedDBStorageAdapter = () => {
           type: "application/json",
         });
 
+        // Generate unique name for each metadata file
+        const uniqueName = `metadata_${nanoid()}.json`;
```
han4wluc, line 210: sorry I missed the `storeMetadata` file. what is it, what does it do? what is file content?
is this only for waveform or all files?

738NGX: only for waveform, store the wave data like this:

```json
{
  "data": [
    0,
    32,
    ......
  ],
  "duration": 233.73333333333332,
  "sampleRate": 44100,
  "channels": 2
}
```

src/components/waveformVisualizer/waveformVisualizer.handlers.js
738NGX: now reduced transform times

src/deps/indexedDBStorageAdapter.js
han4wluc: ok. some notes:
* round duration.
* instead of `data`, use something more specific. is it frequency or like waveformData or the correct name for it.

also document it somewhere in code related to waveform.

src/deps/indexedDBStorageAdapter.js
```diff
@@ -205,9 +206,12 @@ export const createIndexedDBStorageAdapter = () => {
           type: "application/json",
         });
 
+        // Generate unique name for each metadata file
```
han4wluc, line 209: I see how `storeMetadata` works now. basically is mimicking the File js object.

* json -> string -> blob

basically it is for storing json files.

ok for now. but let me think if there should be a more general interface.

han4wluc (approved):

src/deps/fileManager.js
```diff
@@ -220,9 +230,13 @@ export const createFileManager = ({ storageAdapter, fontManager }) => {
...
-        // Convert byte values (0-255) back to normalized values (0-1) for waveform data
-        if (data && data.data && Array.isArray(data.data)) {
-          data.data = data.data.map((value) => value / 255);
+        // Convert byte values (0-255) back to normalized values (0-1) for waveform amplitudes
+        // This is the reverse of the compression done during storage
+        if (data && data.amplitudes && Array.isArray(data.amplitudes)) {
+          data.amplitudes = data.amplitudes.map((value) => value / 255);
+          // TODO: Change rtgl-waveform to accept raw waveform data format directly
+          // Now also provide 'data' field for compatibility with rtgl-waveform components
+          data.data = data.amplitudes;
```
han4wluc, line 239: no need this for backward compatibility thing.
remove `data.data`, update it in `rtgl-waveform` to use `amplitudes`.

otherwise it is just old code that becomes tech debt, we want to clean it up so code remains more clean.

PR #231, opened 2025-09-18
feat bundle

738NGX:


src/deps/bundleService.js
```diff
@@ -0,0 +1,255 @@
...
+  const exportProject = async (projectData, files = {}) => {
+    return await createBundle(projectData, files);
+  };
+
+  /**
+   * Imports a project from a bundle file
+   * @param {File|ArrayBuffer} file - Bundle file to import
+   * @returns {Object} Imported project data with blob URLs
+   */
+  const importProject = async (file) => {
```
han4wluc: oh this is not needed.
bundling and exporting is only for distribution use, not for importing back into a project.
we will probably do some more things in future during bundling such as removing information that is unneeded for the distributed. such as the names of images, etc...

src/pages/settings/settings.view.yaml
```diff
@@ -34,5 +42,11 @@ template:
               - rtgl-text s=sm col=muted: "Current Version:"
               - rtgl-text s=sm: "${appVersion}"
             - rtgl-button#checkUpdateButton variant=pr: Check for Updates
+          - rtgl-view d=v g=md mt=xl:
+            - rtgl-text s=lg: Project Bundle
+            - rtgl-text s=sm col=muted mb=md: Export and import project bundles
+            - rtgl-view d=h g=md ah=c:
+              - rtgl-button#createBundleButton variant=pr: Create Bundle
+              - 'rvn-drag-drop#importDragDrop .uploadText="Import Bundle" .acceptedFileTypes=${[".vnbundle"]}':
```
han4wluc: no need for import. can remove.

han4wluc (approved):

PR #232, opened 2025-09-21
update tauri dev config

han4wluc:


src-tauri/tauri.conf.json
```diff
@@ -6,8 +6,6 @@
   "build": {
     "frontendDist": "../_site",
     "devUrl": "http://localhost:3001",
-    "beforeDevCommand": "bun run watch:tauri",
-    "beforeBuildCommand": "bun run build:tauri",
```
han4wluc, line 10: can keep those 2 separate.

for watch:
so in 1 terminal session you run `bun run watch:tauri`
and other session run the client.

for build:
`"tauri:build": "bun run build:tauri && tauri build"` already runs `bun run build:tauri``

src-tauri/tauri.conf.json
```diff
@@ -41,11 +41,11 @@
       "icons/icon.icns",
       "icons/icon.ico"
     ],
-    "createUpdaterArtifacts": "v1Compatible"
+    "createUpdaterArtifacts": true
```
han4wluc, line 44: we don't need v1 compatibility because we don't have any old versions built with v1

src-tauri/tauri.conf.json
```diff
@@ -41,11 +41,11 @@
...
   },
   "plugins": {
     "updater": {
-      "pubkey": "dW50cnVzdGVkIGNvbW1lbnQ6IG1pbmlzaWduIHB1YmxpYyBrZXk6IDlDRjJDOEUyNEY4OEZFQjYKUldTMi9vaFA0c2p5bkNadWk5NGNpbkhEQ1d0cnIwRHV3RFhCSDB2cE5XNU1YbCszeVdzTDBUNVoK",
+      "pubkey": "dW50cnVzdGVkIGNvbW1lbnQ6IG1pbmlzaWduIHB1YmxpYyBrZXk6IDM4ODJGRjUyRTg5NDBFRDcKUldUWERwVG9VditDT0gyQTNEWmV3Q0RqUElnUW54aGs0NDhGNGdtL0JvZW00SEdZeDh5ZFl4UnkK",
```
han4wluc, line 48: I've generated a new private, public key pair. private one added to `env.dev`

this one we use only for local development.
for production deployment will generate and add it so only CI environment has access to it.

package.json
```diff
@@ -22,18 +22,15 @@
...
     "build:tauri": "rm -rf _site && mkdir _site && cp -rf static/* _site/ && rtgl ui build-svg && rtgl fe build -s src/setup.tauri.js",
-    "watch": "rtgl fe watch -s src/setup.web.js",
+    "watch:web": "rtgl fe watch -s src/setup.web.js",
     "watch:tauri": "rtgl fe watch -s src/setup.tauri.js",
     "serve": "bunx serve _site -p 3001",
     "lint": "bunx prettier --check src",
     "lint:fix": "bunx prettier --check --write src",
     "prepare": "husky",
     "tauri": "tauri",
-    "tauri:dev": "tauri dev",
```
han4wluc, line 36: make it more explicit what platform to run

PR #238, opened 2025-09-23
Feat version

738NGX:
Only basic version save and delete implemented, export bundle will implement in next pr

https://github.com/user-attachments/assets/867dec51-5392-47ab-ba83-63f057c49a06

### DIFF
--- a/src/pages/versions/versions.handlers.js
+++ b/src/pages/versions/versions.handlers.js
@@ -1,3 +1,82 @@
-export const handleAfterMount = async (deps) => {};
+import { nanoid } from "nanoid";
 
-export const handleDataChanged = (e, deps) => {};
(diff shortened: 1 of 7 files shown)


han4wluc (approved):
LGTM

src/pages/versions/versions.handlers.js
```diff
@@ -0,0 +1,141 @@
...
+
+    // Get current action count from repository
+    const allEvents = repository.getAllEvents();
+    const currentActionId = allEvents.length;
+
+    // Create simple version object
+    const newVersion = {
+      id: nanoid(),
+      name: formData.name,
+      actionId: currentActionId,
```
han4wluc, line 54: this is not actionId. this should be called more like `actionIndex`

PR #245, opened 2025-09-30
Feat initial scene

738NGX:
Now the init scene use red text color to display.

Export works well, can correctly get the id.

Also replaced some `store.getState()` in old codes into set and select function.


han4wluc (approved):
LGTM

PR #247, opened 2025-09-30
feat projects icon

738NGX:


src/deps/projectsService.js
```diff
@@ -1,34 +1,26 @@
...
-
-    return null;
+  const loadProjectIcon = async (projectId, iconFileId) => {
+    if (!iconFileId) return null;
+    const fileManager = await fileManagerFactory.getByProject(projectId);
+    console.log(projectId, iconFileId, fileManager); // ok, fileManager is not undefined
+    const { iconUrl } = await fileManager.getFileContent({
+      fileId: iconFileId,
+    });
+    console.log("iconUrl", iconUrl); // error here, iconUrl is undefined
```
738NGX: here is the problem, the logic is same as file-image, but dont know why cannot get image url

han4wluc: anything from here?
```js
    const result = await fileManager.getFileContent({
      fileId: iconFileId,
    });
    console.log(result)
```

738NGX: image

OK i knew, should be `url` instead of `iconUrl`, now it can work well

han4wluc:
hmm, I think the image should be at left of the title, not right. do you have opinion on this?

han4wluc (approved):

738NGX:
> hmm, I think the image should be at left of the title, not right. do you have opinion on this?

A long time ago when I first started working on this interface, I tried to put it on the left, but encountered some problems so I put the placeholder on the right. I can't remember the specific problem now. Merge this PR first and take a look at it later.

PR #268, opened 2025-10-08
feat group view items delete

738NGX:
Main logic is the same as file explorer

### DIFF
--- a/src/components/groupResourcesView/groupResourcesView.handlers.js
+++ b/src/components/groupResourcesView/groupResourcesView.handlers.js
@@ -290,8 +290,7 @@ export const handleCloseContextMenu = (deps) => {
 };
 
 export const handleContextMenuClickItem = async (deps, payload) => {
-  const { store, render, dispatchEvent, repositoryFactory, router, props } =
-    deps;
+  const { store, render, dispatchEvent, props } = deps;
   const detail = payload._event.detail;
   // Extract the actual item (rtgl-dropdown-menu wraps it)
   const item = detail.item;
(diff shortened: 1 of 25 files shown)


src/components/groupResourcesView/groupResourcesView.view.yaml
```diff
@@ -265,7 +286,7 @@ template:
           - rtgl-view w=f h=f av=c ah=c:
               - rtgl-text s=lg c=mu-fg: No data
 
-              - rtgl-view h=64:
+      - rtgl-dropdown-menu#context-menu .items=dropdownMenu.items ?open=${dropdownMenu.isOpen} x=${dropdownMenu.x} y=${dropdownMenu.y}:
```
han4wluc: put dropdown at the very bottom of the view.

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -267,3 +267,75 @@ export const handleAfterMount = async (deps) => {
...
+      );
+    }
+
+    const repositoryState = repository.getState();
+    const targetData = repositoryState[repositoryTarget];
+    const currentItem =
+      targetData && targetData.items ? targetData.items[itemId] : null;
+
+    if (currentItem) {
+      repository.addAction({
```
han4wluc: component should not call repository.
we want component to dispatchEvent.
and in the page to handle the event and call repository.

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -267,3 +267,75 @@ export const handleAfterMount = async (deps) => {
...
+    deps;
+  const detail = payload._event.detail;
+  // Extract the actual item (rtgl-dropdown-menu wraps it)
+  const item = detail.item;
+  const dropdownMenu = store.selectDropdownMenu();
+  const itemId = dropdownMenu.targetItemId;
+
+  // Only handle delete actions
+  if (item && item.value === "delete-item") {
+    const repositoryTarget = props.repositoryTarget;
```
han4wluc: it should be no need to pass `repositoryTarget` as we already have `resourceType.`

src/components/groupResourcesView/groupResourcesView.handlers.js
738NGX: ok, because `fileExplorer` is also component and call repository

If we want to page call repository, is that means need add a new handler function to all pages? I think it might be more complicated

han4wluc: not more complicated but just more repetition.

conclusion:
- correct architecture is for components not to use repository. they only deal with view.
- yes lets add handler to all pages
- if we don't do this, we will do more wrong architecture which will be even harder to fix in future  

don't worry about file explorer for now will fix another time.

han4wluc (approved):

PR #269, opened 2025-10-08
Fix 1

738NGX:
- fixed add character
- fixed font load


src/components/fontPreview/fontPreview.handlers.js
```diff
@@ -1,25 +1 @@
...
-        fontName: fontFamily,
-        fileId: fileId,
-      };
-      await loadFontFile(fontItem);
-    } catch (error) {
-      console.warn(`Failed to load font: ${fontFamily}`, error);
-    }
-  }
-};
+export const handleAfterMount = async () => {};
```
han4wluc: can remove entirely if is empty function.

src/pages/fonts/fonts.handlers.js
```diff
@@ -4,11 +4,39 @@ import { toFlatItems } from "../../deps/repository.js";
...
 
 export const handleAfterMount = async (deps) => {
-  const { store, repositoryFactory, router, render } = deps;
+  const { store, repositoryFactory, router, render, fileManagerFactory } = deps;
   const { p } = router.getPayload();
   const repository = await repositoryFactory.getByProject(p);
   const { fonts } = repository.getState();
   store.setItems(fonts);
+
+  // Load all fonts before rendering to ensure proper preview display
```
han4wluc: could you give motivation why you decided to move font loading from component to the page.
and what are the tradeoffs?
what was the original root cause of the bug?

738NGX: The root problem should be the font load in component is behind the init render, so it show the fallback font. Only double click it recalled render, then show the correct one.

If automatically recall render the time cannot be estimated when load logic in component, so do that in page. This only need one call after all data prepared

738NGX: completed, in the component, call `render` after `handleAfterMount` can work

han4wluc (approved):

PR #275, opened 2025-10-08
fix waveform display

738NGX:
The solution is only render when data exists


han4wluc (approved):
is `waveformData` with values 0 to 255 or 0 to 1?

738NGX:
> is `waveformData` with values 0 to 255 or 0 to 1?

0-255

PR #283, opened 2025-10-09
refactor file explorer

738NGX:


738NGX:
`file explorer click can change the page's selectingItemId` fully implemented

src/components/baseFileExplorer/baseFileExplorer.handlers.js
```diff
@@ -326,6 +326,16 @@ export const handleItemClick = (deps, payload) => {
       },
     }),
   );
+
+  dispatchEvent(
```
han4wluc: this should not be necessary. it is same as `click-item` one above.

han4wluc (approved):

PR #304, opened 2025-10-15
feat init scene display

738NGX:
image

image

### DIFF
--- a/src/components/whiteboard/whiteboard.view.yaml
+++ b/src/components/whiteboard/whiteboard.view.yaml
@@ -89,13 +89,13 @@ template:
   - 'rtgl-view#container w=f h=f pos=rel bgc=bg cur=${containerCursor} style="overflow: hidden; min-height: 400px;"':
     - 'rtgl-view#canvas w=f h=f pos=rel style="transform: translate(${panX}px, ${panY}px) scale(${zoomLevel}); transform-origin: 0 0; width: 0px; height: 0px;"':
       - $for item in items:
-        - 'rtgl-view#item-${item.id} w=300 h=60 d=h ah=e av=c cur=${itemCursor} pos=abs style="left: ${item.x}px; top: ${item.y}px; z-index: 10; user-select: none;"':
+        - 'rtgl-view#item-${item.id} w=300 h=60 d=h ah=e av=c pos=abs style="left: ${item.x}px; top: ${item.y}px; z-index: 10; user-select: none;"':
           - $if item.isInit == true:
             - rtgl-view d=h av=c:
               - rtgl-view p=md bgc=ac br=md bw=sm bc=${item.borderColor}:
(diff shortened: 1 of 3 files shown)


src/components/whiteboard/whiteboard.view.yaml
```diff
@@ -89,9 +89,14 @@ template:
   - 'rtgl-view#container w=f h=f pos=rel bgc=bg cur=${containerCursor} style="overflow: hidden; min-height: 400px;"':
     - 'rtgl-view#canvas w=f h=f pos=rel style="transform: translate(${panX}px, ${panY}px) scale(${zoomLevel}); transform-origin: 0 0; width: 0px; height: 0px;"':
       - $for item in items:
-        - 'rtgl-view#item-${item.id} pos=abs w=120 h=60 bgc=mu bw=sm bc=${item.borderColor} br=md cur=${itemCursor} style="left: ${item.x}px; top: ${item.y}px; z-index: 10; user-select: none;"':
-          - rtgl-view w=f h=f av=c ah=c p=md:
-            - rtgl-text s=sm c=${item.textColor}: "${item.name}"
+        - 'rtgl-view#item-${item.id} w=300 h=60 d=h ah=e av=c cur=${itemCursor} pos=abs style="left: ${item.x}px; top: ${item.y}px; z-index: 10; user-select: none;"':
```
han4wluc: changing the `w=300` will also change how coordinates work.
now the origin of the rectangle has been shifted to empty space in the left. if put mouse in the empty space in the left, it can be dragged.
this should not be the case. it should not impact the coordinates

738NGX: In fact, can also not set w to let the width be automatically controlled. The cost is that each time set the initial scene, there will be a visual shift to the right of the main body. If can accept this,

han4wluc: of course can't accept that.
you need to satisfy both:
- no shifting when item is initial or not
- coordinates should be based on the normal item without the left initial box.

738NGX: A simple change is keep the most, only modified the cur setting, and then prevent the drag happen when `itemX + 180 > mouseInCanvasX * zoomLevel`

738NGX: The small problem is that the label of `initial scene` cannot be clicked or dragged. If necessary, can add another layer of conditional support

han4wluc: this solves issue on UI side, but the coordinates and origin position are still wrong.
need to use different approach such as:
- use position absolute
- negative margin (not ideal)
- draw it outside the element. calculate again the correct coordinate where it should be displayed

738NGX: now use position absolute, completed

han4wluc (approved):

PR #308, opened 2025-10-16
move back projects button

738NGX:
image

### DIFF
--- a/src/pages/project/project.handlers.js
+++ b/src/pages/project/project.handlers.js
@@ -64,3 +64,12 @@ export const handleFormExtraEvent = async (deps) => {
     console.error("Project image upload failed:", error);
   }
 };
+
+export const handleBackToProjects = async (deps) => {
+  const { subject } = deps;
+
+  // Navigate back to projects page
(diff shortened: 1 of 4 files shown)


han4wluc:
one issue is there is too much padding in the left for the button. anyone we can remove or decrease that left padding?

han4wluc:
> one issue is there is too much padding in the left for the button. anyone we can remove or decrease that left padding?

remove padding in the outermost container
create a container below the button. in this container add some little padding

738NGX:
> > one issue is there is too much padding in the left for the button. anyone we can remove or decrease that left padding?
> 
> remove padding in the outermost container create a container below the button. in this container add some little padding

image

Changed

han4wluc:
ok. lastly too much vertical distance between button and bottom content. need to decrease this this distance.

han4wluc (approved):

738NGX:
> ok. lastly too much vertical distance between button and bottom content. need to decrease this this distance.

image

Ok, now this is the shortest vertical distance, button only set mt and lower container only set ph

PR #315, opened 2025-10-17
feat scene grid bg

738NGX:
image

### DIFF
--- a/src/components/whiteboard/whiteboard.store.js
+++ b/src/components/whiteboard/whiteboard.store.js
@@ -173,6 +173,16 @@ export const selectViewData = ({ state, props }) => {
     const canvasGridSize = 20;
     const visualGridSize = canvasGridSize * zoomLevel;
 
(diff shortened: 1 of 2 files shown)


han4wluc:
looks good.
but there is when zoom out. the dots become too dense. and start giving out 密集恐惧症. what could we do to alleviate that?

738NGX:
React Flow directly limits the minimum zoom ratio so that it does not appear too dense. However, when it is reduced to a certain ratio, Twinery will reorganize the grid according to a certain ratio (for example, combining a 4x4 grid into one grid). I think Twinery may be a better approach.

han4wluc:
yes, can try that, so we decrease quantity of dots to make less dense. but I want to keep dots like React Flow, not use line grids like Twinery.

738NGX:
https://github.com/user-attachments/assets/c495c261-6fcd-402c-9533-1b5efe5ddd72


Completed

han4wluc (approved):
LGTM!

PR #320, opened 2025-10-20
feat scenes arrows update

738NGX:
The origin implement is directly change the item dom data, but it seems hard for arrows. My plan is to dispatchEvent when dragging and scenes page listening that, then update the whiteboard props and force re-render, however, the `item-position-updating` cannot be listened

### DIFF
--- a/src/components/whiteboard/whiteboard.handlers.js
+++ b/src/components/whiteboard/whiteboard.handlers.js
@@ -249,6 +249,7 @@ export const handleItemContextMenu = (deps, payload) => {
 
 export const handleWindowMouseMove = (deps, payload) => {
   const { store, getRefIds, render } = deps;
+  console.log(store.selectIsDragging());
 
   if (store.selectIsPanning()) {
     // Handle panning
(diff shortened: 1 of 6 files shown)


src/components/whiteboard/whiteboard.handlers.js
```diff
@@ -300,12 +301,17 @@ export const handleWindowMouseMove = (deps, payload) => {
...
     const snappedY = Math.round(constrainedY / 5) * 5;
 
-    // Update the element position directly for immediate visual feedback
-    const itemElement = getRefIds()[`item-${dragItemId}`];
-    if (itemElement && itemElement.elm) {
-      itemElement.elm.style.left = snappedX + "px";
-      itemElement.elm.style.top = snappedY + "px";
-    }
+    // Dispatch real-time position update to parent (scenes page)
+    console.log(snappedX, snappedY);
```
738NGX: This log has the value

src/pages/scenes/scenes.handlers.js
```diff
@@ -147,6 +147,25 @@ export const selectSelectedItemId = ({ state }) => {
   return state.selectedItemId;
 };
 
+export const handleWhiteboardItemPositionUpdating = async (deps, payload) => {
+  console.log("handleWhiteboardItemPositionChanged called!");
+  const { store, render } = deps;
+  const { itemId, x, y } = payload._event.detail;
+
+  console.log("Updating item position:", { itemId, x, y });
```
738NGX: This log has no output, the func has not been called

src/components/whiteboard/whiteboard.handlers.js
```diff
@@ -249,6 +249,7 @@ export const handleItemContextMenu = (deps, payload) => {
 
 export const handleWindowMouseMove = (deps, payload) => {
   const { store, getRefIds, render } = deps;
```
han4wluc: forgot to get `dispatchEvent` from `deps`.
without that, it uses global `dispatchEvent` which is not the one we want.

738NGX: Ok, now solved and can work correctly

https://github.com/user-attachments/assets/f4af8872-7c11-46f6-978c-6e6ecac1d3d3

han4wluc (approved):

PR #321, opened 2025-10-20
update esbuilded main.js

738NGX:


han4wluc:
file is too big, I don't want to put in git.
also does not say how to build this. the next person may not be able to build it in a repetitive way.

maybe you can have it generated from a script. and we add the script into the source code.

PR #322, opened 2025-10-20
update versions page

738NGX:
image

image

Note: 
- no 3 dots & download icon
- How to set the padding or gap size in button? how to change the text color for button (and wonder if the `Download Zip` need set grey color)
- Selecting download path haven't been implemented, will do in next pr

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,7 +4,7 @@
     "": {
       "dependencies": {
         "@rettangoli/fe": "0.0.10",
-        "@rettangoli/ui": "0.1.12",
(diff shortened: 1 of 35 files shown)


han4wluc:
ask Claude Code to create the svg you need in the svg folder. it is usually very good at it.
buttons, we only have the ones in rettangoli-ui, pick the best you can find.

738NGX:
image

Done. The gap between icon and text in button seems to wide. maybe can let rtgl-button support gap adjust in the future

han4wluc (approved):

PR #328, opened 2025-10-22
feat esbuild script

738NGX:
Operations:

```
cd static/bundle
npm install
npm run build
```

will generate `bundle.min.js`, will use that.


scripts/main.js
```diff
@@ -204,4 +204,6 @@ const init = async () => {
 
 };
 
-await init();
+(async () => {
```
han4wluc: this is same a just

`init()`

static/bundle/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
han4wluc: no, we don't want extra package.json. and especially don't want it under static. because static folder is deployed and accessible to public internet.

instead. you can try update https://github.com/yuusoft-org/routevn-creator-client/blob/main/scripts/build.sh  for building this.

static/bundle/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "bundle",
+  "version": "1.0.0",
+  "description": "",
+  "main": "main.js",
+  "scripts": {
+    "build": "esbuild main.js --bundle --format=iife --minify --outfile=bundle.min.js"
```
han4wluc: I think you can use `--format=esm`. so that you can keep using `<script src="./main.js" type="module">`

static/bundle/package.json
738NGX: Well, if there's no cdn or extra package.json, it means should manually download all packages in scripts？

han4wluc: you move the original `main.js` into the `scripts` folder. 
I think you can reuse the root `project.json`

738NGX: ok, will try

738NGX:
Now completed

han4wluc (approved):

static/bundle/.gitignore
```diff
@@ -0,0 +1 @@
+main.js
```
han4wluc: cannot put this in static because it will be deployed to public.
use the .gitignore in root folder

han4wluc (approved):

PR #330, opened 2025-10-22
feat icon validation

738NGX:


src/pages/characters/characters.handlers.js
```diff
@@ -220,6 +222,10 @@ export const handleFormExtraEvent = async (deps) => {
 
   const file = files[0];
 
+  if (!(await checkIconDimensions(file, globalUI))) {
```
han4wluc: I don't like passing globalUI like this. if we want to pass globalUI like that, need to put it in deps.

do this instead:

```js
const { isValid, message } = validateIconDimensions(file);
if (!isValid) {
     globalUI.showAlert({
        message: `Image must be square. Current size: ${dimensions.width}x${dimensions.height}`,
        title: "Error",
      });
     return
}

```

src/utils/fileProcessors.js
```diff
@@ -20,6 +20,38 @@ export const getImageDimensions = (file) => {
   });
 };
 
+export const checkIconDimensions = async (file, globalUI) => {
+  const dimensions = await getImageDimensions(file);
+
+  if (!dimensions) {
+    await globalUI.showAlert({
```
han4wluc: `showAlert` I think is not async since does not require confirmation.

src/pages/characters/characters.handlers.js
738NGX: now using this

han4wluc (approved):

PR #333, opened 2025-10-23
refactor uploader

738NGX:
image

Changelog:
- Feat upload/add button, the same as the old ones in group view
- the old ones in group view will have full width, and only appears when folder is empty
- drag & drop haven't been implemented, will do in next pr

### DIFF
--- a/src/components/groupResourcesView/groupResourcesView.handlers.js
+++ b/src/components/groupResourcesView/groupResourcesView.handlers.js
@@ -1,3 +1,8 @@
+import {
+  getAcceptAttribute,
+  isFileTypeAccepted,
+} from "../../utils/fileTypeUtils.js";
+
 export const handleSearchInput = (deps, payload) => {
   const { dispatchEvent } = deps;
   const searchQuery = payload._event.detail.value || "";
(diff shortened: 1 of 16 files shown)


han4wluc:
- button add upload/add icon. make it `v=pr`. so white is more easier to see.

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -49,6 +54,60 @@ export const handleItemDoubleClick = (deps, payload) => {
...
+  input.accept = getAcceptAttribute(props.acceptedFileTypes);
+  input.multiple = true;
+  input.onchange = (e) => {
+    if (e.target && e.target.files) {
+      const validFiles = Array.from(e.target.files).filter((file) =>
+        isFileTypeAccepted(file, props.acceptedFileTypes),
+      );
+
+      if (validFiles.length > 0) {
+        handleDragDropFileSelected(deps, {
```
han4wluc: calling handler from handler is still a bit weird not very clean.

try this:

* create a normal function.
* in both handlers call this function.

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -49,6 +54,60 @@ export const handleItemDoubleClick = (deps, payload) => {
...
+  } else if (resourceType === "colors") {
+    handleAddColorClick(deps, newPayload);
+  } else if (resourceType === "typography") {
+    handleAddTypographyClick(deps, newPayload);
+  } else if (resourceType === "layouts") {
+    handleAddLayoutClick(deps, newPayload);
+  } else if (resourceType === "transforms") {
+    handleAddTransformClick(deps, newPayload);
+  } else if (resourceType === "animations") {
+    handleAddAnimationClick(deps, newPayload);
```
han4wluc: all those functions. they are more or less doing the same thing I think. only the event name is diffrent.

try to see if you can:
* collapse all of them into 1 function
* can use same event name for all pages. check if this is possible on causes any issues.

src/components/groupResourcesView/groupResourcesView.handlers.js
738NGX: About this, the final step need the `dispatchEvent` & `fontManager` deps, so i think the common funtion will still be a handler, so just copy the `handleDragDropFileSelected` logic to `handleUploadButtonClick` instead of call that

src/components/groupResourcesView/groupResourcesView.handlers.js
738NGX: These all merged to use `handleAddButtonClick`

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -80,141 +136,35 @@ export const handleDragDropFileSelected = async (deps, payload) => {
...
-  payload._event.stopPropagation(); // Prevent group click
-  const groupId = payload._event.currentTarget.id.replace(
-    "add-character-button-",
-    "",
-  );
+export const handleAddButtonClick = (deps, payload) => {
+  const { props, dispatchEvent } = deps;
+  const { resourceType } = props;
+  payload._event.stopPropagation();
+  const singularResourceType = resourceType.endsWith("s")
```
han4wluc: which ones is using plurar?
I think can directly rename to singular from the source of it

src/components/groupResourcesView/groupResourcesView.handlers.js
```diff
@@ -80,141 +136,35 @@ export const handleDragDropFileSelected = async (deps, payload) => {
...
+    ? resourceType.slice(0, -1)
+    : resourceType;
+  const groupId = payload._event.currentTarget.id.replace("add-btn-", "");
 
-  // Forward add character click to parent
   dispatchEvent(
-    new CustomEvent("add-character-click", {
+    new CustomEvent(`add-${singularResourceType}-click`, {
```
han4wluc: I mean can we just do ``add-click` ? so we update code in pages as well to listen only to`add-click`

738NGX: seems ok, now have been implemented

han4wluc (approved):

PR #339, opened 2025-10-24
modified drag&drop

738NGX:


738NGX:
text tip now still keeps in center, may modify to mouse position in later pr

han4wluc (approved):

PR #341, opened 2025-10-24
Feat pan mode btn

738NGX:
https://github.com/user-attachments/assets/47dce9a4-bef8-4691-9574-2553d0a0ff3d


han4wluc:
- put the pan button to the left of zoom.
- replace it with a hand icon instead of text
- when is active, update button `v` instead. and also update the hand icon for active (optional)

738NGX:
image

image

Completed

han4wluc (approved):

PR #344, opened 2025-10-27
Fix fs access

738NGX:
Solutions:
- give the dialog `recursive` access, so can get `/**`
- should have `features = ["protocol-asset"]`, or get asset files will meet 403


han4wluc:
using fs API. can user write/read to the `.persistent-scope` file?

han4wluc:
also, have you check if sqlite can read/write to which locations? is it exactly same as fs?

src-tauri/capabilities/default.json
```diff
@@ -37,8 +37,7 @@
       "identifier": "fs:scope",
       "allow": [
         "$APPDATA/**",
```
han4wluc: can this be removed?

src-tauri/tauri.conf.json
```diff
@@ -68,8 +68,7 @@
       "assetProtocol": {
         "enable": true,
         "scope": [
-          "$APPDATA/**",
-          "**"
+          "$APPDATA/**"
```
han4wluc: which asset files do we have under this?
don't we have only the sqlite file under APPDATA which does not need to be accessed by asset?

738NGX:
> using fs API. can user write/read to the `.persistent-scope` file?

In plugin source code, this file has been defined cannot be access

```rs
if let Some(fs_scope) = &fs_scope {
    let _ = fs_scope.forbid_file(&fs_scope_state_path);
}
```

and also be binary coded, if directly modify can raise error

```rs
let _ = f.write_all(&bincode::serialize(&scope).map_err(Error::from)?);
```

> also, have you check if sqlite can read/write to which locations? is it exactly same as fs?

- the `app.db`: in `$APPDATA/**`, always can access
- the `repository.db` in project path: same as fs, rely on `.persistent-scope`

han4wluc:
We need to have some test cases to have confidence in this:
create a html page for testing. can be just pure html page and write javascript directly inside this html page.
then, do all the things that we need to validate such as:
- reading & writing files/folders inside and outside the scope
- use sql to read/write to location inside and outside the scope.
based on the result, update the text in the UI. to show if test has passed or not.

this will also help up in future to make sure these tests always pass

738NGX:
> We need to have some test cases to have confidence in this: create a html page for testing. can be just pure html page and write javascript directly inside this html page. then, do all the things that we need to validate such as:
> 
> * reading & writing files/folders inside and outside the scope
> * use sql to read/write to location inside and outside the scope.
>   based on the result, update the text in the UI. to show if test has passed or not.
> 
> this will also help up in future to make sure these tests always pass

image

have done the test. seems sqlite still can access, may need fix that

han4wluc:
> > We need to have some test cases to have confidence in this: create a html page for testing. can be just pure html page and write javascript directly inside this html page. then, do all the things that we need to validate such as:
> > 
> > * reading & writing files/folders inside and outside the scope
> > * use sql to read/write to location inside and outside the scope.
> >   based on the result, update the text in the UI. to show if test has passed or not.
> > 
> > this will also help up in future to make sure these tests always pass
> 
> image
> have done the test. seems sqlite still can access, may need fix that

❌ and ✅ should depend on whether the test is performing as expected or not. not whether it can access file or not. also check for access to `.persistent-scope` file.

738NGX:
Test page has been added

han4wluc (approved):

src-tauri/capabilities/default.json
738NGX: yes "$APPDATA/**" have been all removed and can still work

PR #347, opened 2025-10-29
feat message pack

738NGX:
image

now values in sqlite are all coded

### DIFF
--- a/src/deps/keyValueStore.js
+++ b/src/deps/keyValueStore.js
@@ -9,7 +9,7 @@ export const createKeyValueStore = async () => {
     await db.execute(`
       CREATE TABLE IF NOT EXISTS ItemTable (
         key TEXT PRIMARY KEY,
-        value TEXT
+        value BLOB
       )
     `);
   };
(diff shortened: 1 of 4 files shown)


src/deps/tauriRepositoryAdapter.js
```diff
@@ -133,19 +154,27 @@ export const createTauriSQLiteRepositoryAdapter = async (projectPath) => {
...
       },
       set: async (key, value) => {
-        const jsonValue = JSON.stringify(value);
+        // Encode to MessagePack, then convert to Base64 string for storage
+        const binaryData = encode(value);
+        let binaryString = "";
+        for (let i = 0; i < binaryData.length; i++) {
+          binaryString += String.fromCharCode(binaryData[i]);
+        }
+        const base64Value = btoa(binaryString);
```
han4wluc: wait we don't need this stuff.
you have to change column type from TEXT to BLOB.
then can store binary directly. don't convert back to text.
yes, it will be a breaking change for all previous stored data.

738NGX: done, now directly store blob, but the sql plugin returns a json array when request blob data, so now still need a `json.parse`

src/deps/keyValueStore.js
```diff
@@ -20,20 +21,20 @@ export const createKeyValueStore = async () => {
     );
 
     if (result && result.length > 0) {
-      try {
-        return JSON.parse(result[0].value);
-      } catch {
-        return result[0].value;
-      }
+      // Tauri SQL returns JSON array as string, parse it first
+      const numberArray = JSON.parse(result[0].value);
```
han4wluc, line 25: do we have anything else in `result` that is closer to binary?

han4wluc: what is value of `value` ? is it like this? `"[104,101,108,108,111]"`?

han4wluc: it seems unfortuantely the tauri sql lib automatically converts to json/string 😢 .
can spend a little time to look at source code of `@tauri-apps/plugin-sql` , if no easy solution we keep like this.

738NGX: > what is value of `value` ? is it like this? `"[104,101,108,108,111]"`?

yes

```
=== DATA DEBUG ===
keyValueStore.js:25 Raw value: [145,132,162,105,100,181,109,68,79,81,82,55,118,50,77,88,111,74,120,68,65,45,119,101,73,82,105,171,112,114,111,106,101,99,116,80,97,116,104,181,68,58,92,68,111,99,117,109,101,110,116,115,92,86,78,92,84,101,115,116,49,169,99,114,101,97,116,101,100,65,116,207,0,0,1,154,46,62,250,253,172,108,97,115,116,79,112,101,110,101,100,65,116,192]
keyValueStore.js:26 Type: string
keyValueStore.js:27 Constructor: String
keyValueStore.js:28 Array.isArray: false
keyValueStore.js:29 Length: 335
keyValueStore.js:32 First 10 elements: (10) ['[', '1', '4', '5', ',', '1', '3', '2', ',', '1']0: "["1: "1"2: "4"3: "5"4: ","5: "1"6: "3"7: "2"8: ","9: "1"length: 10[[Prototype]]: Array(0)
keyValueStore.js:34 
=== END DEBUG ===
```

han4wluc:
closing this. due to the serialization from rust to js. loses performance benefit of message pack.

PR #348, opened 2025-10-29
Modify bundle service

738NGX:
Note:

- changed all `.vnbundle` to `bin`
- structure is `[version(1)] [indexLength(4)] [reserved(11)] [index(JSON)] [assets...] [instructions(JSON)]`, commented in `bundleService.js`
- removed `extractBundle` because it has never been used
- updates on main.js to read the new bundle file

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -14,12 +14,30 @@ import RouteGraphics, {
 
 async function parseVNBundle(arrayBuffer) {
   const uint8View = new Uint8Array(arrayBuffer);
-  const headerSize = 9 + Number(new DataView(uint8View.buffer, 1, 8).getBigUint64(0));
-  const index = JSON.parse(new TextDecoder().decode(uint8View.subarray(9, headerSize)));
+  const dataView = new DataView(arrayBuffer);
+
+  // Read version (byte 0) - not used for now but good to have
+  const version = uint8View[0];
+  if (version !== 1) {
+    throw new Error(`Unsupported bundle version: ${version}`);
+  }
+
+  // Read index length (bytes 1-4, big-endian)
+  const indexLength = dataView.getUint32(1, false);
+
+  // Fixed header size is 16 bytes
+  const headerSize = 16;
+
+  // Read index (starts after 16-byte header)
+  const index = JSON.parse(new TextDecoder().decode(uint8View.subarray(headerSize, headerSize + indexLength)));
   const assets = {};
   let instructions = null;
+
+  // Data block starts after header and index
+  const dataBlockOffset = headerSize + indexLength;
+
   for (const [id, metadata] of Object.entries(index)) {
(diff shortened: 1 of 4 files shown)


scripts/main.js
```diff
@@ -14,12 +14,30 @@ import RouteGraphics, {
 
 async function parseVNBundle(arrayBuffer) {
   const uint8View = new Uint8Array(arrayBuffer);
-  const headerSize = 9 + Number(new DataView(uint8View.buffer, 1, 8).getBigUint64(0));
-  const index = JSON.parse(new TextDecoder().decode(uint8View.subarray(9, headerSize)));
+  const dataView = new DataView(arrayBuffer);
+
+  // Read version (byte 0) - not used for now but good to have
```
han4wluc:
```suggestion
  // Read engine version (byte 0) - Different versions numbers means incompatible formats
```

scripts/main.js
```diff
@@ -14,12 +14,30 @@ import RouteGraphics, {
 
 async function parseVNBundle(arrayBuffer) {
   const uint8View = new Uint8Array(arrayBuffer);
-  const headerSize = 9 + Number(new DataView(uint8View.buffer, 1, 8).getBigUint64(0));
-  const index = JSON.parse(new TextDecoder().decode(uint8View.subarray(9, headerSize)));
+  const dataView = new DataView(arrayBuffer);
```
han4wluc, line 16: how come we have both `uint8View` and `dataView` , can we use only one?

src/deps/bundleService.js
```diff
@@ -64,114 +66,63 @@ export const createBundleService = () => {
...
-      versionBuffer.length + indexLengthBuffer.length,
-    );
-
-    // Add all assets and instructions
-    for (const arrayBuffer of arrayBuffers) {
-      finalBundle.set(
-        new Uint8Array(arrayBuffer.responseArrayBuffer),
-        arrayBuffer.start + headerSize,
-      );
+    // Get version from repository app table, fallback to 1
```
han4wluc: dont' fall back. throw error if version is not found.

han4wluc: actually, do we need this? creator version we don't really care about it during bundling and unbundling.

738NGX: so what's the version in bundle? engine or creator? if engine, how get that?

han4wluc: for the bundle, it is engine version. hardcoded to 1 for now.

scripts/main.js
738NGX: now all use `dataView`

src/deps/bundleService.js
738NGX: this done

han4wluc (approved):

PR #352, opened 2025-10-30
feat scenes page posotion save

738NGX:
- [x] Scene Map, when going to another page and then coming back, should still show the same position. We probably need to save info locally.

### DIFF
--- a/src/pages/scenes/scenes.handlers.js
+++ b/src/pages/scenes/scenes.handlers.js
@@ -1,4 +1,5 @@
 import { nanoid } from "nanoid";
+import { createUserConfig } from "../../deps/userConfig";
 
 /**
  *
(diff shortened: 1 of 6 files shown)


src/pages/scenes/scenes.handlers.js
```diff
@@ -81,6 +82,20 @@ export const handleAfterMount = async (deps) => {
   // Initialize whiteboard with scene items only
   store.setWhiteboardItems(sceneItems);
 
+  // Restore viewport state from userConfig
+  const userConfig = createUserConfig();
+  const savedZoomLevel = userConfig.get("scenesView.zoomLevel");
```
han4wluc: lets call it `scenesMap`

738NGX: renamed

han4wluc (approved):

PR #353, opened 2025-10-30
make start different

738NGX:
- [x] Scenes Map page. Start rectangle looks too similar to a node. make it a bit more different (a user confused and though Start is a complete separate node that can be double clicked). make Start not selectable, and that it will not highlight.

image

### DIFF
--- a/src/components/whiteboard/whiteboard.view.yaml
+++ b/src/components/whiteboard/whiteboard.view.yaml
@@ -92,8 +92,8 @@ template:
       - 'rtgl-view#canvas w=f h=f pos=rel style="transform: translate(${panX}px, ${panY}px) scale(${zoomLevel}); transform-origin: 0 0; width: 0px; height: 0px;"':
           - $for item in items:
               - $if item.isInit == true:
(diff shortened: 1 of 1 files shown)


han4wluc:
`start` text is too crowded, make circle a bit bigger

han4wluc (approved):

738NGX:
> `start` text is too crowded, make circle a bit bigger

because current is `br=xl`, if let it bigger it will become rounded rectangle, do you think should keep current br and let it be rounded rectangle, or use custom style to force it circle?

han4wluc:
> > `start` text is too crowded, make circle a bit bigger
> 
> because current is `br=xl`, if let it bigger it will become rounded rectangle, do you think should keep current br and let it be rounded rectangle, or use custom style to force it circle?

can use `br=f`

738NGX:
> > > `start` text is too crowded, make circle a bit bigger
> > 
> > 
> > because current is `br=xl`, if let it bigger it will become rounded rectangle, do you think should keep current br and let it be rounded rectangle, or use custom style to force it circle?
> 
> can use `br=f`

image

set wh=50

han4wluc (approved):

PR #364, opened 2025-11-18
Feat/update to 0.0.7 route graphics

ggbb0711:
This is still a WIP. I'm still having problems with loading sprite, and there are still console.logs to clean up

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -107,15 +107,14 @@ export const layoutTreeStructureToRenderState = (
           fontFamily: fontItem?.fontFamily || "sans-serif",
           fontWeight: typography.fontWeight || "normal",
           fill: colorItem?.hex || "white",
-          lineHeight: typography.lineHeight || 1.5,
-          fontFileId: fontItem?.fileId ? `file:${fontItem.fileId}` : undefined,
+          lineHeight: typography.lineHeight || 1.2,
         };
       } else {
         // Use default settings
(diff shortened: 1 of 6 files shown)


src/utils/index.js
```diff
@@ -141,11 +140,9 @@ export const layoutTreeStructureToRenderState = (
             fontFamily: hoverFontItem?.fontFamily || "sans-serif",
             fontWeight: hoverTypography.fontWeight || "normal",
             fill: hoverColorItem?.hex || "white",
-            lineHeight:
-              (hoverTypography.lineHeight || 1.5) *
-              (hoverTypography.fontSize || 24),
+            lineHeight: hoverTypography.lineHeight || 1.5,
```
han4wluc: default to 1.2

han4wluc: prevouis 1.5 *24 was wrong.

han4wluc (commented):
generally looking good.
let me know if you found anything about the sprite loading issue.

src/utils/index.js
ggbb0711: The texture.from(src) always return null for the sprite. And I check the Assets.cache, it has the key but the value is null, so I think the problem is because of the path or something?

han4wluc: yeah, I think something along these lines... check for when we're loading the assets.
needs some debugging.

PR #365, opened 2025-11-19
Add lic to installer

Prabesh002:
Add lisc to bundler

from [Task 006](https://github.com/RouteVN/routevn-creator-client/blob/main/tasks/TASK/000/TASK-006.md)

image


han4wluc (approved):
LGTM

PR #366, opened 2025-11-20
update to schema of api

Prabesh002:
[TASK-004](https://github.com/RouteVN/routevn-creator-client/blob/main/tasks/TASK/000/TASK-004.md)
[TASK-006](https://github.com/RouteVN/routevn-creator-client/blob/main/tasks/TASK/000/TASK-006.md)
[TASK-007](https://github.com/RouteVN/routevn-creator-client/blob/main/tasks/TASK/000/TASK-007.md)

### DIFF
--- a/src/deps/tauriUpdater.js
+++ b/src/deps/tauriUpdater.js
@@ -96,6 +96,15 @@ const createUpdater = (globalUI) => {
         }
       });
     }, 5000);
+
+    const SIX_HOURS_IN_MS = 6 * 60 * 60 * 1000;
+    setInterval(() => {
+      checkForUpdates(true).then((info) => {
+        if (info) {
+          console.log("Update available on interval check:", info);
+        }
+      });
+    }, SIX_HOURS_IN_MS);
   };
 
   return {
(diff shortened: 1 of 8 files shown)


src/deps/tauriUpdater.js
```diff
@@ -90,12 +90,21 @@ const createUpdater = (globalUI) => {
...
-      checkForUpdates(false).then((info) => {
+      checkForUpdates(true).then((info) => {
         if (info) {
           console.log("Update available on startup:", info);
         }
       });
     }, 5000);
+
+    const SIX_HOURS_IN_MS = 6 * 60 * 60 * 1000;
+    setInterval(() => {
```
han4wluc: instead of using setInterval over long time which is pretty unreliable.

lets store a last checkUpdateTime using our app db key value.

Then, we have a setIntravel for every 10min or so, and compare with the last checkUpdateTime value. and if this is larger than our threshold go check the update, and update the db.

src/deps/tauriUpdater.js
han4wluc: lets test this. 

give me a test json with a real test .exe with the test sign key.

we will make version like 99999. so it will always ask to update.

I'll update this json to server so we can all use it to test.

han4wluc (approved):

738NGX (approved):

PR #369, opened 2025-11-24
Feat/web support

Prabesh002:
This is a work in progress and to track bugs on  the web

### Bugs
- [x] Any time you upload multiple files it will be lost on the middle panel (when inspecting, it will show on the side panel) happens on most pages
- [x] Cannot delete the files if the first bug occurs (mostly on the character sprites page, but haven't tested others)
- [ ] Can't preview the scene 
- [ ] Can't export bundle

Will add more the more I find out


Prabesh002:
So it was not the multiple file that was the issue but the files with spaces, it works fine on the tauri one but if I upload a file named 

`me and chair` it will throw `Failed to load image: Error: File not found: me`

image

It will show properly but missing on the middle page

I think it was because ` const fileId = file.name.replace(/\.[^/.]+$/, "");` we're doing this (I didn't notice this on the adaptor) so will update if this is not the culprit

Prabesh002:
This was the reason, it is fixed

PR #377, opened 2025-12-02
increase default zoom

738NGX:
image


han4wluc (approved):
do we need any changes in default coordinate, or current one is good?

738NGX:
> do we need any changes in default coordinate, or current one is good?

use (120, 90) instead of (0, 0) to insure center

image

han4wluc (approved):

PR #396, opened 2025-12-04
fix typography form default value reset

738NGX:
force form rerender

### DIFF
--- a/src/pages/typography/typography.view.yaml
+++ b/src/pages/typography/typography.view.yaml
@@ -90,8 +90,7 @@ template:
               - rtgl-form#typography-form key=${formKey} .defaultValues=dialogDefaultValues .form=dialogForm w=f:
           - rtgl-view flex=1 p=md bwl=xs d=v:
               - rtgl-text s=lg fw=b mb=md: Preview
-              - $if isDialogOpen:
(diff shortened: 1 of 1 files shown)


src/pages/typography/typography.view.yaml
han4wluc: we already have one here. which is the component that needs to be re-rendered?

738NGX: this only re-render the preview but not form, so the form won't be updated. maybe can remove the one here and keep the upper one, should also can re-render both form and preview. i dont know what's the key reason, but in fact it's the only way i can find to make it work correctly, and should be the simplest change

src/pages/typography/typography.view.yaml
```diff
@@ -84,6 +84,7 @@ template:
                           - rtgl-text s=md c=mu-fg: No selection
 
   - rtgl-dialog#add-typography-dialog ?open=${isDialogOpen}:
+    - $if isDialogOpen:
       - rtgl-view slot=content d=h g=md w=800:
           - rtgl-view flex=1:
               - rtgl-form#typography-form key=${formKey} .defaultValues=dialogDefaultValues .form=dialogForm w=f:
```
han4wluc, line 90: `key=${isDialogOpen}`

try this first. if does not work. ok... use existing one.

han4wluc (approved):

src/pages/typography/typography.view.yaml
han4wluc: ok. then jsut remove this extra end condition

PR #408, opened 2025-12-05
Fix/border event not override elements

ggbb0711:


han4wluc (approved):
with this. will preview for hover and click events still work?

ggbb0711:
Sorry, for the late comment, but to answer your question, click and hover are still working. It just when you select an element the border will take priority over the other element-> Making the click and hover not work only when the element is selected. Is that alright?

han4wluc:
> Sorry, for the late comment, but to answer your question, click and hover are still working. It just when you select an element the border will take priority over the other element-> Making the click and hover not work only when the element is selected. Is that alright?

sorry. miseed this comment and PR.

to be honest I don't know, would need to test it....
do you think this is good, does it have any issues?

han4wluc (approved):

PR #409, opened 2025-12-08
Feat/web init

Prabesh002:
It's basically the same as the tauri one but I just switched to index db


han4wluc (approved):
can give it a try

PR #412, opened 2025-12-08
l10n setup

han4wluc:
- **update**
- **implmeent localization**


han4wluc:
this was an attemp to add localization.
however there are several issues:
- keys are long
- work only for specific dialogue content structure
- i'm a bit uneasy of using variables directly. would to have a key and it turned into variable in the route-engine constructRenderState

decision:
- will put this on hold, and proceed with other things
  - this means that when we do introduce localization, there will be most likely a breaking change.

PR #414, opened 2025-12-08
add code signing key

han4wluc:


han4wluc:
can run the script on windows.
will build tauri on linux, and signon windows.

PR #415, opened 2025-12-09
Client Visual test

Prabesh002:
For now, I've added a simple but effective way to capture these screenshots by using an iframe now it will just render a test template that contains an iframe pointing to our running dev server. This allows to generate reference candidate images of actual application pages. We can later update the rettangoli library to support the navigate and the URL support so we can navigate and support some complex actions like typing on text fields

Currently need to build and run the bun run serve on separate terminal and then run the test in different terminal.

Need to discuss about more complex scenario later on


vt/specs/projects/projects.yaml
```diff
@@ -0,0 +1,10 @@
+---
+title: Initial Project Page Load
+description: Verifies that the projects page loads correctly.
+steps:
+#TODO : to make something in vt so it can navigate to a path before taking screenshot
+# then maybe it would be possible to test the client..
+ #- navigate: /projects
```
han4wluc, line 7: ok. naming of these lets try to follow the naming from playwright functions so we don't have to make up everything ourself.

vt/templates/default.html
```diff
@@ -0,0 +1,30 @@
...
+        left: 0;
+        width: 100%;
+        height: 100%;
+        border: none;
+      }
+    </style>
+  </head>
+  <body>
+    <!--- needs to be changed later, kept simpler for just screenshot -->
+    <iframe src="http://localhost:3002/projects"></iframe>
```
han4wluc, line 28: I didn't think of putting iframe here but seems smart clean solution. lets use this for now.

package.json
```diff
@@ -31,7 +31,10 @@
     "build:tauri": "./scripts/build.sh tauri",
     "watch:web": "rtgl fe watch -s src/setup.web.js",
     "watch:tauri": "rtgl fe watch -s src/setup.tauri.js",
-    "serve": "bunx serve _site -p 3001",
+    "vt:generate": "bun run build:web && rtgl vt generate --skip-screenshots",
+    "vt:report": "bun run build:web && rtgl vt generate && rtgl vt report",
+    "vt:accept": "rtgl vt accept",
+    "serve": "bunx serve _site -p 3002",
```
han4wluc: lets keep serve for non vt scenario. for vt maybe need to add a `vt:serve`

rettangoli.config.yaml
```diff
@@ -17,6 +17,10 @@ vt:
   path: 'vt'
   name: "Rettangoli UI Visual Tests"
   description: "Visual testing for Rettangoli UI components"
+  port: 3002
+  serve:
+    - from: /
```
han4wluc: i don't remember. do we really have this or is it hallucinated?

vt/templates/default.html
han4wluc: just need to be caseful iframe may not be 100% same as the original. sometimes add some styling which I think you already remove all of it.

han4wluc (approved):

rettangoli.config.yaml
Prabesh002: seems hallucinated, didnt notice it. Will remove

vt/specs/projects/projects.yaml
Prabesh002: ok sure, will look at docs to see if anything matches

vt/templates/default.html
Prabesh002: yes removed mostly everything that may caused weird styling so should not be an issue!

PR #420, opened 2025-12-11
Feat/add back animations

ggbb0711:


han4wluc (commented):
please also post a screenshot or screen recording of the testing.
so it is easier to review.

for this feature, I think what needs to be tested is 
- whehter there is any other way to break it. like adding extreme values, or empty or what etc...
- preview works correctly

ggbb0711:
Note:  Value type: Relative(Keyframes with alpha symbol) mean that the animation will add on the value from the previous keyframe instead of setting them
Pos X testing:

https://github.com/user-attachments/assets/c267f8b8-2f1e-4161-97ff-5c546f56f928

Pos Y testing: 

https://github.com/user-attachments/assets/dd9f7598-1071-4fb1-b725-2944c51b0f63

Alpha testing: 

https://github.com/user-attachments/assets/306f5370-4e3c-4041-8c92-847b79e9b9f6

Scale X testing: 

https://github.com/user-attachments/assets/127e6cdb-bb4b-464d-8922-e52530afa45b

Scale Y testing: 

https://github.com/user-attachments/assets/bb74c21e-19a6-4667-a0d5-681069bdd567

Combined testing: 

https://github.com/user-attachments/assets/74c19ffa-0e33-49db-93da-af94359cd92a

han4wluc (approved):

PR #430, opened 2025-12-22
Visual tests for client

Prabesh002:
This is a WIP. Also not sure if I have to manually change the ui version on the HTML files, since when I updated it wasn't updated so I updated them all.

Also the page is proper but there's still an issue for candidate page, need to fix that

image

The problem is that if we try to load any candidate from the candidate folder it will result in this type of page but if we go back and try to use the app as normal it will work in the browser so I'm a bit stuck on how to solve this so this needs to be discussed today

### DIFF
--- a/vt/specs/projects/projects.yaml
+++ b/vt/specs/projects/projects.yaml
@@ -1,7 +1,3 @@
 ---
 title: Initial Project Page Load
-steps:
-  #- goto https://rettangoli.dev
-  #- wait 2000
-  #- screenshot
 ---
--- a/vt/specs/projects/projects.yaml
+++ b/vt/specs/projects/projects.yaml
@@ -2,9 +2,7 @@
 title: Initial Project Page Load
 description: Verifies that the projects page loads correctly.
 steps:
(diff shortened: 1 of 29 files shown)


vt/specs/projects/projects.yaml
```diff
@@ -2,9 +2,5 @@
 title: Initial Project Page Load
 description: Verifies that the projects page loads correctly.
 steps:
-#TODO : to make something in vt so it can navigate to a path before taking screenshot
-# then maybe it would be possible to test the client..
- #- navigate: /projects
-  - wait: 2000
   - screenshot
```
ggbb0711: The vt has already take a screenshot before the steps, so the screenshot here is unnecessary

Prabesh002: Oh for this, i was testing the navigation so I forgot to remove it

han4wluc:
add a url property somewhere like in the spec yaml content matter, or in the cli command itself.

ggbb0711:
LGTM

PR #438, opened 2026-01-06
Bug Report In Client

Prabesh002:
#### Since on the tasks, you will see the videos name they will be this

[scene_character_bug]
https://github.com/user-attachments/assets/2d8a30e4-8ffb-4123-8f18-4969150d3569

[dual_character_bug]
https://github.com/user-attachments/assets/d8c114a2-ae29-47db-902e-bdaa6acaa6bb

[unresponsive_scene]
https://github.com/user-attachments/assets/85eacfb9-a18f-4e11-86f0-57c992acc591

**Some issues block from creating a proper visual novel, we should give high priority to scene related first**

### DIFF
--- a/tasks/TASK/000/TASK-038.md
+++ b/tasks/TASK/000/TASK-038.md
@@ -1,0 +1,76 @@
+---
+title: Bug Report
+status: todo
+priority: high
+assignee: anyone
+---
+
+# Description
+
+# Feedback & Issues Report
+
+## General UI / UX Issues
+
+- The input to enter the width and height for the layout container are still annoying, and the popups are still not repoisitioning correctly when first click
+- The animation editor is confusing(it combine both left/right click to activate poup, it is better to be consistence with this)
+- Somehow animation preview is not working
+- Don't know if this is my computer problem, but whenever I click to other tabs for a while and return back to the client, it freeze for a solid 20 seconds or so.
+- To delete something you need to click right click on it for a popup then choose delete, it might be better for the ux to also include an 'x' or a trashcan icon as well as right clicking to delete something
+- On the left layout, whenever there is no folder or file, it has "No items, Click to create a folder" text appears, it might be beeter to include a button here or highlight the text more to make people less confuse on how to create a folder
+- Having to go to another meny to change the character position on the screen is really anoying
+- Should add prevent default as it still functions like a browser
+
+---
+
+## Layout Editor Issues & Suggestions
+
+- In the layout editor, adding an option to stretch and shrink the element with your mouse without clicking the menu on the left would be nice
+- In the layouts editor, we can't click and drag them directly, we must click on the left menu to interact with them
+- After sometime of not focusing on the client and return back to the client, I got the this._cancelResize error. And the layouts don't show me anything
+
+---
+
+## Scene Editor Issues
+
+- Scene Editor, when first create a new scene, it give me a black screen with only a plus symbol on the left, I can't click on the tabs at the top, because there were none, so I click on the plus symbol and it create another tab
+- In the scene editor, if a line has too many text, it push the preview to the right.
+- Scene editor take a huge amount of ram on my laptop, and it is really slow
+- Scene editor, choice is not forcing people to choose. You can click on the textbox to ignore the choice completely
+- Scene editor, can't choose which section to start the preview from
+- Choice is still not working
+- In scene editor the "Add Scene" button at the bottom center doesn't turn on the popup immediately, but it need to be pressed on the editor zomming map to turn on the popup. Need to turn on the popup immediately
+- In scene editor, the arrow only correct between scenes whenever there is an arrow from one scene to another
+
+---
+
+## Bugs
+
+1) when adding character on scene it's buggy and when deleting the character it's not properly deleting (or showing delete option where the character is not there) and not showing the option where the character is there  
+video name -> `scene_character_bug`
+
+2) Character immediately not updated on the scene page when i add on the same line with existing character  
+video name -> `dual_character_bug`
+
+3) When removing items it's not immediately updated on state page properly and sometimes evn on the section page
+
+4) Unresponsive Scene layout, it's very buggy as shown in the below video  
+video name -> `unresponsive_scene`
+
+5) If audio resource is removed from asset and if it's present on that line then it will not trigger that line and show blank  
+error -> `Uncaught TypeError: s is not iterable`
+
+---
+
+## UX Problems
+
+1) On character we need to be able to layer pictures like expressions or other things which is not supported in the client  
+2) On the sounds page, when playing audio, can make the slider more visible so it's intent is clear  
+3) User can benefit from keyboard actions, like delete to delete, f2 to rename, and other stuff which is not here in the creator
+
+---
+
+## Confused / Suggestions
+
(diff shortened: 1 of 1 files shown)


tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+
+1) On character we need to be able to layer pictures like expressions or other things which is not supported in the client  
+2) On the sounds page, when playing audio, can make the slider more visible so it's intent is clear  
+3) User can benefit from keyboard actions, like delete to delete, f2 to rename, and other stuff which is not here in the creator
+
+---
+
+## Confused / Suggestions
+
+- Adding multiple character on the same line causes to hide the previous character (if the character is same) need info on this, not sure if a bug or intended
```
han4wluc, line 81: yes this is intended. discussed on discord

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+- In the layouts editor, we can't click and drag them directly, we must click on the left menu to interact with them
+- After sometime of not focusing on the client and return back to the client, I got the this._cancelResize error. And the layouts don't show me anything
+
+---
+
+## Scene Editor Issues
+
+- Scene Editor, when first create a new scene, it give me a black screen with only a plus symbol on the left, I can't click on the tabs at the top, because there were none, so I click on the plus symbol and it create another tab
+- In the scene editor, if a line has too many text, it push the preview to the right.
+- Scene editor take a huge amount of ram on my laptop, and it is really slow
```
han4wluc: is this for the development version or the download production version?
how much ram is it taking?

ggbb0711: This is for the development, I think it might be a local problem.

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+# Description
+
+# Feedback & Issues Report
+
+## General UI / UX Issues
+
+- The input to enter the width and height for the layout container are still annoying, and the popups are still not repoisitioning correctly when first click
+- The animation editor is confusing(it combine both left/right click to activate poup, it is better to be consistence with this)
+- Somehow animation preview is not working
+- Don't know if this is my computer problem, but whenever I click to other tabs for a while and return back to the client, it freeze for a solid 20 seconds or so.
```
han4wluc: need to find the root cause for this.
if you encoutner it, other users will also encounter it

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+- To delete something you need to click right click on it for a popup then choose delete, it might be better for the ux to also include an 'x' or a trashcan icon as well as right clicking to delete something
```
han4wluc: right click to delete is ok, as long as is consistent all around.

problem with x is:
- more ugly, takes space
- can be misclicked, so will need to add a confirmation dialog.

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+
+- In the layout editor, adding an option to stretch and shrink the element with your mouse without clicking the menu on the left would be nice
+- In the layouts editor, we can't click and drag them directly, we must click on the left menu to interact with them
+- After sometime of not focusing on the client and return back to the client, I got the this._cancelResize error. And the layouts don't show me anything
+
+---
+
+## Scene Editor Issues
+
+- Scene Editor, when first create a new scene, it give me a black screen with only a plus symbol on the left, I can't click on the tabs at the top, because there were none, so I click on the plus symbol and it create another tab
```
han4wluc: are the lines there? there should be like 30 empty lines.

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+- In the scene editor, if a line has too many text, it push the preview to the right.
+- Scene editor take a huge amount of ram on my laptop, and it is really slow
+- Scene editor, choice is not forcing people to choose. You can click on the textbox to ignore the choice completely
+- Scene editor, can't choose which section to start the preview from
+- Choice is still not working
```
han4wluc: which part of choice is not working?

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+- After sometime of not focusing on the client and return back to the client, I got the this._cancelResize error. And the layouts don't show me anything
+
+---
+
+## Scene Editor Issues
+
+- Scene Editor, when first create a new scene, it give me a black screen with only a plus symbol on the left, I can't click on the tabs at the top, because there were none, so I click on the plus symbol and it create another tab
+- In the scene editor, if a line has too many text, it push the preview to the right.
+- Scene editor take a huge amount of ram on my laptop, and it is really slow
+- Scene editor, choice is not forcing people to choose. You can click on the textbox to ignore the choice completely
```
han4wluc: solution:
in route-engine in selectRenderState addChoice function. add an overlay that will prevent click.

tasks/TASK/000/TASK-038.md
```diff
@@ -0,0 +1,83 @@
...
+- Scene editor, can't choose which section to start the preview from
+- Choice is still not working
+- In scene editor the "Add Scene" button at the bottom center doesn't turn on the popup immediately, but it need to be pressed on the editor zomming map to turn on the popup. Need to turn on the popup immediately
+- In scene editor, the arrow only correct between scenes whenever there is an arrow from one scene to another
```
han4wluc: sorry, didn't get this one.

tasks/TASK/000/TASK-038.md
ggbb0711: no, the lines weren't there

tasks/TASK/000/TASK-038.md
ggbb0711: I meant the arrow only connect between scenes

tasks/TASK/000/TASK-038.md
ggbb0711: I have checked again, and it seem to be a white-space issue, it only breaks with white-space, so If we want to change it we can probably change the `word-breaks` props

tasks/TASK/000/TASK-038.md
ggbb0711: I click on the choice to direct to a new scene but it didn't transition to the new scene.

tasks/TASK/000/TASK-038.md
Prabesh002: This one, can be solved through keyboard actions instead I believe

tasks/TASK/000/TASK-038.md
ggbb0711: Where is this exactly, and when is it? Because I saw in some videos you used to allow 2 characters before?

Prabesh002: It's for the same character only

ggbb0711: why though, don't most visual novel have multiple character on the screens?

han4wluc: he meant adding multiple instance of same characer.
it is common to have character A and character B on the same screen. it is uncommon to have 2 instance of character A on the screen.
anyways this is solved already. can support 2 instance of character A now.

PR #443, opened 2026-01-07
Fix/fix animation

ggbb0711:
I have updated the default `resetState` to make the routeGraphic use updateRect instead of addRect. The problem here is that the white rectangle properties will not be reset to the center when we open up a new animation or add a new animtation. Only when we play a new preview does it update the animation.

https://github.com/user-attachments/assets/aef6ed90-9428-4dd1-8ad4-4a131e61954e


src/pages/tweens/tweens.constants.js
```diff
@@ -0,0 +1,33 @@
+const CANVAS_WIDTH = 1920;
```
han4wluc, line 1: rename to `tweens.constatns.js` to be consistent with other files.

han4wluc (approved):

src/pages/tweens/tweens.store.js
```diff
@@ -1,4 +1,5 @@
 import { toFlatGroups, toFlatItems } from "insieme";
+import { resetState } from "./constants";
```
han4wluc, line 2: @NghiaTT200000 you forgot to update this.

PR #454, opened 2026-01-09
Fix Presentation state and character ID

Prabesh002:
- Fix Presentation State panel not updating after changes (adding/removing characters, navigating lines)
 - Add skipRender flag to prevent cursor reset while typing in line editor
 - Update route engine for character id duplication
 

 ### Presentation State UI fix:

 The State panel was showing stale/inconsistent data because handleRenderCanvas wasn't calling render() after updating presentationState in the store. Added render() call to sync the UI.

### To verify this

  - Add a character on line 2, verify it appears in State panel
  - Go to line 3-4, verify character persists in State panel
  - Delete character on line 5, verify it disappears from State panel
  - Go back to line 2-4, verify character shows correctly

Or any line, they are just for reference

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "jsdom": "^26.1.0",
     "jszip": "^3.10.1",
     "nanoid": "^5.1.5",
-    "route-engine-js": "0.1.14",
+    "route-engine-js": "^0.1.15",
     "route-graphics": "0.0.24",
     "rxjs": "^7.8.2",
     "snabbdom": "^3.6.2",
--- a/src/pages/sceneEditor/sceneEditor.handlers.js
+++ b/src/pages/sceneEditor/sceneEditor.handlers.js
@@ -244,8 +244,9 @@ export const handleEditorDataChanged = async (deps, payload) => {
     content,
   });
 
-  // Render the scene immediately with the updated content
-  // Trigger debounced canvas render
+  // Trigger debounced canvas render with skipRender flag.
+  // skipRender prevents full UI re-render which would reset cursor position while typing.
+  // Typing only updates dialogue content, not presentationState, so State panel doesn't need to update.
   subject.dispatch("sceneEditor.renderCanvas", { skipRender: true });
 };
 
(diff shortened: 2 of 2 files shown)


han4wluc:
update PR title and description to say actually what task/ problem are you solving with this PR.

package.json
```diff
@@ -19,7 +19,7 @@
     "jsdom": "^26.1.0",
     "jszip": "^3.10.1",
     "nanoid": "^5.1.5",
-    "route-engine-js": "0.1.14",
+    "route-engine-js": "^0.1.15",
```
han4wluc, line 22: this is for solving character id duplication thing right?

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -1069,9 +1069,12 @@ export const handleUpdateDialogueContent = async (deps, payload) => {
 };
 
 // Handler for debounced canvas rendering
-async function handleRenderCanvas(deps) {
-  const { store, graphicsService } = deps;
+async function handleRenderCanvas(deps, payload) {
+  const { store, graphicsService, render } = deps;
   await renderSceneState(store, graphicsService);
+  if (!payload?.skipRender) {
```
han4wluc, line 1076: you need to explain why we need this.

this one is easy to fix 1 bug and introduce another bug.

package.json
Prabesh002: yes

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: The render() call is required to update the Presentation State panel UI after presentationState is updated in the store. Without it, the State panel shows stale data when navigating between lines or adding/removing actions

The sceneEditor.renderCanvas event is dispatched from two contexts:

  1. Action changes (add/remove character, navigate lines, etc.) → needs render() to update State panel
  2. Typing in line editor (handleEditorDataChanged) → does NOT need render() because typing doesn't change presentationState

but If we always call render(), the line editor re-renders on every keystroke (debounced 50ms), which resets the cursor position to the beginning.

han4wluc: ok. seems fine. but please do this:
- update PR title and PR description to be more descriptive. 
  - so you're doing 2 separate things in this PR. characterId fix and this.
- you must put this in the code comments, for why we skip render

Prabesh002: done

han4wluc (approved):

PR #458, opened 2026-01-10
Update Scene editor and Presentation State panel to use events

Prabesh002:
#### Previous Attempt and Why It Was Removed

Calling render() on the parent sceneEditor was not ok because:
  1. It re-renders all child components including linesEditor
  2. The contenteditable DOM gets recreated, losing cursor position
  3. This happens asynchronously (50ms debounce), making it impossible to preserve cursor state ( I did not notice this yesterday)

Using the event approach because it's already used in the project

### DIFF
--- a/src/components/systemActions/systemActions.store.js
+++ b/src/components/systemActions/systemActions.store.js
@@ -86,7 +86,6 @@ export const setLocalPresentationState = (state, presentationState) => {
   state.localPresentationState = presentationState;
 };
 
-
 export const selectRepositoryState = ({ state }) => {
   return state.repositoryState;
 };
@@ -132,7 +131,8 @@ export const setMode = (state, payload) => {
 // Moved from sceneEditor.store.js - now returns object instead of array
 export const selectActionsData = ({ props, state }) => {
   const { actions } = props;
-  const presentationState = state.localPresentationState ?? props.presentationState ?? {};
+  const presentationState =
+    state.localPresentationState ?? props.presentationState ?? {};
 
   if (!actions) {
     return {};
--- a/src/components/systemActions/systemActions.store.js
+++ b/src/components/systemActions/systemActions.store.js
@@ -14,6 +14,7 @@ export const createInitialState = () => ({
     ],
   },
   repositoryState: {}, // Add this - default to empty object
+  localPresentationState: null,
 });
 
 export const selectViewData = ({ state, props, attrs }) => {
@@ -81,6 +82,11 @@ export const selectViewData = ({ state, props, attrs }) => {
   };
 };
 
+export const setLocalPresentationState = (state, presentationState) => {
+  state.localPresentationState = presentationState;
+};
+
+
 export const selectRepositoryState = ({ state }) => {
   return state.repositoryState;
 };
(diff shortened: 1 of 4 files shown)


src/components/systemActions/systemActions.store.js
```diff
@@ -125,7 +130,9 @@ export const setMode = (state, payload) => {
 
 // Moved from sceneEditor.store.js - now returns object instead of array
 export const selectActionsData = ({ props, state }) => {
-  const { actions, presentationState = {} } = props;
+  const { actions } = props;
+  const presentationState =
+    state.localPresentationState ?? props.presentationState ?? {};
```
han4wluc: no. this is bad.
adding 2 sources for presentationState would be confusing and lead to inconsistences. alway get this from 1 source.

han4wluc:
in theory calling render in the parent scenesEditor does not automatically trigger render in linesEditor. but I'm not sure, kind of forgot about this, need to check to make sure.

make sure to run the latest version of `rtgl`

but the render in linesEditor happens in:
- https://github.com/RouteVN/routevn-creator-client/blob/268ccd64eb766080abb0e74bf0ec02c44f95ee09/src/pages/sceneEditor/sceneEditor.handlers.js#L677
- https://github.com/RouteVN/routevn-creator-client/blob/268ccd64eb766080abb0e74bf0ec02c44f95ee09/src/pages/sceneEditor/sceneEditor.handlers.js#L667

src/components/systemActions/systemActions.store.js
Prabesh002: used only one source

Prabesh002:
> in theory calling render in the parent scenesEditor does not automatically trigger render in linesEditor. but I'm not sure, kind of forgot about this, need to check to make sure.
> 
> make sure to run the latest version of `rtgl`
> 
> but the render in linesEditor happens in:
> 
>     * https://github.com/RouteVN/routevn-creator-client/blob/268ccd64eb766080abb0e74bf0ec02c44f95ee09/src/pages/sceneEditor/sceneEditor.handlers.js#L677
> 
>     * https://github.com/RouteVN/routevn-creator-client/blob/268ccd64eb766080abb0e74bf0ec02c44f95ee09/src/pages/sceneEditor/sceneEditor.handlers.js#L667

You might be right in theory but I spent hours trying render() approaches and they all caused the cursor bug. Even if it's not automatic propagation, something in the render chain was interfering with contenteditable state which I'm not really sure what's causing it, that's why I asked you about the event based approach


I did try to render just the systemActions element only  but it still had stale props so I just used the events one.

Every render()-based approach either broke the cursor or didn't update the UI properly, so I moved to the event based approach

han4wluc:
context:
- there is nothing wrong with systemActions. data is passed via props, and when props update, it will render. this is how we render everything else.
  - systemActions is also used in layoutEditor. your changes probably broke that page.
- linesEditor is the outlier, it needs different and more careful rendering because of the text input. we cannot render text input because there is cursor with content in it. 



decision:
- do not make special fix for systemActions. instead should fix the special handlings for linesEditor.

PR #459, opened 2026-01-12
Feat/add build to check compiling in workflow

ggbb0711:


.github/workflows/ci.yaml
```diff
@@ -20,3 +20,17 @@ jobs:
 
       - name: Lint
         run: bun run lint
+  build:
```
han4wluc: can merge with the lint job. just run it after the lint step. so we remove duplicate checkout and bun install. is a bit faster.

han4wluc (approved):

PR #467, opened 2026-01-13
Fix/problem with scene editor loading

ggbb0711:
Problem: The scene editor sometime have problem when first loading in. After investigating, I have found out that it was because of loading async operation with route-graphics. I have seperated that part of the code to a different async function. Now the scene editor can load immediately, save for the preview which will have a black screen for a bit before fully loading. 

Note: I'm having problem with using `ctrl  + r` to reload the scene editor page, which cause the line editor to have a black screen. Also, because the core problem is because the graphic loadAssets, we may encounter an error with the preview having a black screen for a while in the future. Which mean we should probably change the loading to only load the nedded assets when changing into a section/scene.

### DIFF
--- a/src/pages/sceneEditor/sceneEditor.handlers.js
+++ b/src/pages/sceneEditor/sceneEditor.handlers.js
@@ -9,6 +9,11 @@ async function createAssetsFromFileIds(
   projectService,
   resources,
 ) {
+  console.log("CREATE ASSETS FROM FILE IDS");
+  console.log("File references: ", fileReferences);
+  console.log("Project service: ", projectService);
+  console.log("Project resources: ", resources);
+  console.log("-----");
   const { sounds, images, fonts = {} } = resources;
   const allItems = Object.entries({
     ...sounds,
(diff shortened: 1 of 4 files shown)


src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -115,21 +138,8 @@ export const handleAfterMount = async (deps) => {
...
-  const assets = await createAssetsFromFileIds(
-    fileReferences,
-    projectService,
-    projectData.resources,
-  );
-  await graphicsService.loadAssets(assets);
-  // don't know why but it needs to be called twice the first time to work...
-  renderSceneState(store, graphicsService);
+  initializeRouteGraphics(canvas, store, projectService, graphicsService);
   render();
```
han4wluc, line 133: I think you can just move this `render` to above the `await graphicsService.init({ canvas: canvas.elm });` to get the same result.

ggbb0711:
Okay, it seem like I was wrong, the error was related to route-graphics. The error was switching from one project to another causing the error: 
`RouteGraphics.js:1093  Uncaught (in promise) Error: [Loader.load] Failed to load http://localhost:3001/project/scene-editor/mLrbsYX2nxgGIR9RQFhsz.
Error: Buffer not found for key: mLrbsYX2nxgGIR9RQFhsz
    at RouteGraphics.js:1093:19670
    at async Promise.all (:3001/project/index 0)
    at async bu.load (RouteGraphics.js:1094:18)
    at async Su._mapLoadToResolve (RouteGraphics.js:1144:7315)
    at async Su.load (RouteGraphics.js:1144:6211)
    at async Promise.all (:3001/project/index 0)
    at async Object.loadAssets (graphicsService.js:76:7)
    at async initializeRouteGraphics (sceneEditor.handlers.js:103:3)`. 
This is related to `route-graphics`, they have an advanced loader that need to be removed whenever RouteGraphics is destroyed.

[Route-graphic pr](https://github.com/RouteVN/route-graphics/pull/185)

https://github.com/user-attachments/assets/1c1df5a2-72f1-4523-960e-c03a36122007

han4wluc (approved):
LGTM. nice finding!

PR #468, opened 2026-01-13
fix character submit action

Prabesh002:


src/components/commandLineCharacters/commandLineCharacters.handlers.js
```diff
@@ -111,9 +111,9 @@ export const handleSubmitClick = (deps) => {
   const selectedCharacters = store.selectSelectedCharacters();
 
   // Only dispatch if there are characters to submit
-  if (selectedCharacters.length === 0) {
-    return;
-  }
+  // if (selectedCharacters.length === 0) {
```
han4wluc: delete code instead of commenting

han4wluc (approved):

PR #473, opened 2026-01-14
Feat/update deletion popup

ggbb0711:
WIP: still need to work on the popup and for other asset type


ggbb0711:
https://github.com/user-attachments/assets/a93ba19d-5ca2-4710-b58e-d38e7b061b2a

src/pages/characters/characters.handlers.js
```diff
@@ -320,6 +325,39 @@ export const handleItemDelete = async (deps, payload) => {
...
+        usage.inProps.scene.push(...sceneUsages);
+        usage.inProps.layout.push(...layoutUsages);
+        usage.count += sceneUsages.length + layoutUsages.length;
+        usage.isUsed = true;
+      }
+    }
+  }
+
+  if (usage.isUsed) {
+    store.showDeleteWarning({ itemId, usage });
```
han4wluc: for dialog use something like this: https://github.com/RouteVN/routevn-creator-client/blob/603debc3c8af8d1701be8094e3bbd991c47e5081/src/pages/projects/projects.handlers.js#L104

so that we don't have to manage all the internal state. should simplify the code by a lot.
can simplify the message to put all into 1 line. for now don't need to mention how many layout and how many scenes. we can may do it later.

src/components/fileExplorer/fileExplorer.handlers.js
```diff
@@ -190,6 +196,94 @@ export const handleFileAction = async (deps, payload) => {
...
     if (currentItem) {
+      let usage = {
+        isUsed: false,
+        inProps: {},
+        count: 0,
+      };
+
+      const resourceType = currentItem.type;
+
+      if (resourceType === "character") {
```
han4wluc, line 197: this is too complicated. isn't this handled by each page already? for example characters is handles in the characters page?
for folders, just check if folder have any children. we don't allow to delete folders with children.

src/pages/tweens/tweens.handlers.js
```diff
@@ -504,6 +509,28 @@ export const handleItemDelete = async (deps, payload) => {
...
+  const layoutUsages = recursivelyCheckResource(
+    state.layouts,
+    itemId,
+    LAYOUT_RESOURCE_KEYS,
+  );
+  const usage = {
+    inProps: { scene: sceneUsages, layout: layoutUsages },
+    isUsed: sceneUsages.length > 0 || layoutUsages.length > 0,
+    count: sceneUsages.length + layoutUsages.length,
+  };
```
han4wluc: would like to simplify this. there is too much stuff in the handler. want to put more into utils.

better way:

```js
const usage = recursivelyCheckResource({ state, itemId, checkTargets: ['layouts', 'scenes'] })

```

src/components/deleteWarningPopup/deleteWarningPopup.view.yaml
```diff
@@ -0,0 +1,44 @@
+elementName: rvn-delete-warning-popup
```
han4wluc: I think no need anymore for this if just use `appService.showToast(message); `

han4wluc (commented):
I think would have been better if you just implmented 1 page, and we quickly reviewed that 1 page first.

src/components/fileExplorer/fileExplorer.handlers.js
ggbb0711: but you can also delete the children in those folder directly

src/pages/characters/characters.handlers.js
```diff
@@ -317,9 +322,37 @@ export const handleDialogAvatarClick = async (deps) => {
...
   const { resourceType, itemId } = payload._event.detail;
 
+  const state = projectService.getState();
+  const character = state.characters.items[itemId];
+
+  let isUsed = false;
+
+  if (character && character.sprites && character.sprites.items) {
+    for (const spriteId of Object.keys(character.sprites.items)) {
+      const usage = recursivelyCheckResource({
```
han4wluc, line 331: would like for `SCENE_RESOURCE_KEYS` and `LAYOUT_RESOURCE_KEYS` to be inside resources so that is not exported. 
from outside we only pass `scenes`, so in the utils we have a mapping from `scenes` to `SCENE_RESOURCE_KEYS`.

src/components/fileExplorer/fileExplorer.handlers.js
han4wluc: ok. so there is 2 ways to delete. one from the group view, and one from file explorer. this one is for file explorer.

not ideal to have this part here, but  I guess it works for now.

han4wluc (approved):

PR #474, opened 2026-01-15
fix export bundle

han4wluc:


tasks/TASK/000/TASK-038.md
```diff
@@ -79,6 +79,8 @@ error -> `Uncaught TypeError: s is not iterable`
 
 - [ ] [High] add adding animation to background, with tween fade in. and then you navigate around different lines, or preview it. the animation runs everytime you click. expected: it should run only once for that line
 
+- [ ] [High] @nellow: bug with dialogue and fade: https://discord.com/channels/1233660682419834920/1461203575123480707/1461203575123480707
```
Prabesh002, line 82: This one, cannot reproduce it on the client anymore. I suspect it's with engine or graphics as jeff changed it's behaviour so waiting for jeff to report findings

PR #488, opened 2026-01-22
fix duplicated content when spammed enter

Prabesh002:
The bug : when splitting a line at cursor position 0 (moving all content to the new line), the old line would show duplicated content instead of being empty

Root Causes:                                                                                                                                               
                                                                                                                                                             
I believe the issue was a race condition between the rendering and the browser's handling of contenteditable elements during changes, I think this is the most likely cause of it.

### DIFF
--- a/src/pages/sceneEditor/sceneEditor.handlers.js
+++ b/src/pages/sceneEditor/sceneEditor.handlers.js
@@ -525,6 +525,22 @@ export const handleSplitLine = async (deps, payload) => {
 
   // Use requestAnimationFrame for focus operations
   requestAnimationFrame(() => {
+    if (leftContent === "" && linesEditorRef && linesEditorRef.elm?.shadowRoot) {
+      const oldLineElement = linesEditorRef.elm.shadowRoot.querySelector(
+        `#line-${lineId}`,
+      );
+      if (oldLineElement) {
+        console.log(
+          `[LE] handleSplitLine | Clearing old line ${lineId} content because split happened at start.`,
+        );
+        oldLineElement.textContent = "";
+
+        // Trigger input event so linesEditor's handlers know about the change
+        const inputEvent = new Event('input', { bubbles: true });
+        oldLineElement.dispatchEvent(inputEvent);
+      }
+    }
+
     if (linesEditorRef) {
       linesEditorRef.elm.transformedHandlers.updateSelectedLine({
         currentLineId: newLineId,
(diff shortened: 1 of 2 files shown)


src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -438,6 +451,10 @@ export const handleSplitLine = async (deps, payload) => {
   // First, update the current line with the left content
   // Only update the dialogue.content, preserve everything else
   const leftContentArray = leftContent ? [{ text: leftContent }] : [];
+  console.log(
```
han4wluc: revemo those logs

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -515,6 +538,24 @@ export const handleSplitLine = async (deps, payload) => {
...
+    ) {
+      const oldLineElement = linesEditorRef.elm.shadowRoot.querySelector(
+        `#line-${lineId}`,
+      );
+      if (oldLineElement) {
+        console.log(
+          `[LE] handleSplitLine | Clearing old line ${lineId} content because split happened at start.`,
+        );
+        oldLineElement.textContent = "";
+        const inputEvent = new Event("input", { bubbles: true });
```
han4wluc, line 573: why do we need to do this?

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -1113,6 +1169,16 @@ export const handleUpdateDialogueContent = async (deps, payload) => {
   const { projectService, store, subject } = deps;
   const { lineId, content } = payload;
 
+  // Skip saving if content is empty (avoids creating { text: "" } structure)
+  const isEmptyContent =
```
han4wluc: if current content is "a". and user cancel it and new content becomes "". will it get saved?

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: to emit a change event, so that old input doesn't hold to the old content

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: ah, currently not. I guess we should only skip if the repository already has empty content. If the repository has "a" 
 and user makes it empty, we need to delete/unset the dialogue

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: removed logs

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: unset properly

han4wluc:
closed in favour of this PR: https://github.com/RouteVN/routevn-creator-client/pull/491

PR #491, opened 2026-01-23
fix lines editor duplicated content on enter

han4wluc:


src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -451,7 +454,7 @@ export const handleSplitLine = async (deps, payload) => {
         },
       },
     });
-  } else if (leftContent) {
+  } else {
```
han4wluc, line 457: when curosr is at beginning and we press enter. `leftContent` would be empty. so neither of the current 2 conditinals would be executed. so data was not saved. this fixes it making sure one of these 2 conditions will always be executed.

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -438,7 +438,10 @@ export const handleSplitLine = async (deps, payload) => {
 
   // First, update the current line with the left content
   // Only update the dialogue.content, preserve everything else
-  const leftContentArray = leftContent ? [{ text: leftContent }] : [];
+  const leftContentArray = leftContent
+    ? [{ text: leftContent }]
+    : [{ text: "" }];
```
Prabesh002, line 443: won't this save to the repository with text : "" instead of just actions with no dialogue?

src/pages/sceneEditor/sceneEditor.handlers.js
```diff
@@ -521,7 +526,10 @@ export const handleSplitLine = async (deps, payload) => {
         currentLineId: newLineId,
       });
 
-      // Also render the linesEditor
+      linesEditorRef.elm.transformedHandlers.forceSyncContentLine({
```
han4wluc, line 529: after debugging and checking the logs. I confirmed that the data is all correct. state is updated and correct. but however for some reason the div contenteditable content is still wrong, it does not update.
conclude that this is a bug in the frontend framework, diff algorithm, for some reason it does not diff it correctly. 
solution: surgically manually update the dom.

src/components/linesEditor/linesEditor.handlers.js
```diff
@@ -867,6 +867,22 @@ export const handleOnInput = (deps, payload) => {
...
+
+  const { lineId } = payload;
+  const lineRef = refIds[`line-${lineId}`];
+  // Check if lineRef exists and has the elm property
+  if (!lineRef || !lineRef.elm) {
+    return;
+  }
+  const lineContent = store.selectLineContent({ lineId });
+  if (lineRef.elm.textContent !== lineContent) {
+    lineRef.elm.textContent = lineContent;
```
han4wluc, line 882: manually updates the dom. with the latest data.

src/pages/sceneEditor/sceneEditor.handlers.js
Prabesh002: also a weird bug occurs but if the line has text as "" the cursor seems to go at the end of the right side

han4wluc: > won't this save to the repository with text : "" instead of just actions with no dialogue?

`scenes.items.${sceneId}.sections.items.${sectionId}.lines.items.${lineId}.actions.dialogue.content`

we update dialogue directly. does not impact other actions.

han4wluc: > also a weird bug occurs but if the line has text as "" the cursor seems to go at the end of the right side

create a full reproduction of this. and we will debug it separatedly.

Prabesh002: https://discord.com/channels/1233660682419834920/1463492068746723339/1464097849720700939

here, is the reproduction. Pulled from this branch

PR #499, opened 2026-01-24
Add preview from selected line and section id

Prabesh002:
https://github.com/user-attachments/assets/dac4f35b-fdea-4f0c-8c1b-436d339bc3e9

### DIFF
--- a/src/components/vnPreview/vnPreview.handlers.js
+++ b/src/components/vnPreview/vnPreview.handlers.js
@@ -43,46 +43,32 @@ export const handleAfterMount = async (deps) => {
   const state = projectService.getState();
   const { canvas } = getRefIds();
   await graphicsService.init({ canvas: canvas.elm });
-  const projectData = constructProjectData(state, {
-    initialSceneId: attrs["scene-id"],
-  });
-  const fileReferences = extractFileIdsFromRenderState(projectData.resources);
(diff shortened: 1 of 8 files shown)


src/components/vnPreview/vnPreview.handlers.js
```diff
@@ -52,4 +52,37 @@ export const handleAfterMount = async (deps) => {
   await graphicsService.initRouteEngine(projectData, {
     handleEffects: true,
   });
+
```
han4wluc, line 46: you see that the below code is very messy? We want to be able to handle this better in a more clean way.

the engine already support a custom initialSectionId. but need add support for `initialLineId`.

It would look like this:

```
const projectDataWithInitial = structuredClone(projectData);
projectDataWithInitial.story.scenes[sceneId].initialSectionId = sectionId;
projectDataWithInitial.story.scenes[sceneId].sections[sectionId].initialLineId = lineId;
await graphicsService.initRouteEngine(projectDataWithInitial);
```



need to implement `initialLineId` in https://github.com/RouteVN/route-engine/blob/64f6ba4b1718fc7c999d2405679cc9634a9a1845/src/stores/system.store.js#L29

Prabesh002: resolved, shortly will open engine PR

han4wluc (approved):

PR #506, opened 2026-01-25
When item is clicked on layout editor make it interactable instead of selecting from file panel

Prabesh002:
https://github.com/user-attachments/assets/060313fc-6153-4f5b-ae98-7b4480ef70b8

currently attempting to select any element will result in the container getting selected, but when selected from file preview, it will work and attempt to click on background will also lead to container getting selected.

- i think in route-graphics, when you click an element, the event fires for that element and then bubbles up to its parents so this might be the cause but I'm not sure


han4wluc:
cannot review as there is no product requirements and specification to refer to.

PR #510, opened 2026-01-25
Fix version not loading properly

Prabesh002:


han4wluc (commented):
weird but ok for now

han4wluc (approved):
weird but ok for now

PR #515, opened 2026-01-27
Update template data

Prabesh002:


static/templates/default/repository.json
```diff
@@ -5,9 +5,17 @@
   },
   "images": {
     "items": {
-      "zZO4u50ZEEt1WIugqEiW4": {
+      "bg_folder": {
```
han4wluc: better keep random generated id. so is more random

han4wluc (approved):

static/templates/default/repository.json
Prabesh002: Not sure if you mean to all or just this folder, for now used random on bg_folder as before

han4wluc: for all

Prabesh002: done

han4wluc (approved):

PR #519, opened 2026-01-28
add findings for scene view UX

Prabesh002:


docs/research/scene-research.md
```diff
@@ -0,0 +1,57 @@
...
+    - Havent tried this as it required installation but it seems like hovering over a node (which represents a note) shows a preview of its content in a pop-up window
+    - A single click on a node opens the corresponding note in a new editor pane.
+
+## What might suit RouteVn?
+
+Since every app has different behaviour, and by looking at the behaviour of these platforms we can do this
+
+
+- When a user hovers over scene nodes, we can highlight it
+-  When a user hovers over a scene node, a small edit button with pencil icon appears on the node. Clicking this button navigates to the scene editor. We also keep the double click functionality, this way the user might know what the action is intended for
```
Jeff-Y-work, line 53: I may prefer showing the edit button if right clicked? Keeping double click is good idea.

docs/research/scene-research.md
```diff
@@ -0,0 +1,57 @@
...
+- For the start circle, i think it's better if we remove it, and instead on the initial scene, we add a visual indicator like "Start" on top of the scene like
```
Jeff-Y-work, line 54: We can even do some simple highlight for users to better locate the start.

Jeff-Y-work (approved):
LGTM

docs/research/scene-research.md
Prabesh002: Hmm, you know how when we hover it shows on corner, my idea was that haha, users may discover it when hovering and know what action it might be

Jeff-Y-work: Right, it probably works fine :D

han4wluc: adding over effect is fine.

we still need single click to select the scene. so it is consistent with the other pages. and we need selection to show more detail stuff in the right panel.

docs/research/scene-research.md
han4wluc: yeah... i agree to improve the indicator. not sure if we have any better alternatives than start text.
plain start text solve the issue, so i'm fine with it. just it may not best looking.

PR #531, opened 2026-02-09
feat/variable page

Jeff-Y-work:
- **Added page for variables and set default system variables**
- **fixed lint**

https://github.com/user-attachments/assets/75e5c7c6-f37b-4357-b260-df93717ac83f

A problem is that when adding variable in new folder, the column doesn't align. I haven't found good solution yet.

Major refactor to previous variables page, such as renaming and deleting unimplemented things. 

Array/object type variable not supported yet, seems a bit different for the interface.

### DIFF
--- a/src/components/groupVariablesView/groupVariablesView.handlers.js
+++ b/src/components/groupVariablesView/groupVariablesView.handlers.js
@@ -117,4 +117,3 @@ export const handleFormActionClick = (deps, payload) => {
     render();
   }
 };
-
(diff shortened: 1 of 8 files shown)


Jeff-Y-work:
Now changed from using table to a custom layout. This makes them align.


https://github.com/user-attachments/assets/0f99f5d1-2983-4b7f-a8df-17a5871bf75c

Also for chosen row, background changes colour.

Jeff-Y-work:
https://github.com/user-attachments/assets/fadd12e0-cf86-4ee3-a944-45a7a340bd09

Selection part able to modify variable.

han4wluc (approved):
ok. looks good. custom layout is good idea. no need for array and object  (is complicated)

PR #535, opened 2026-02-10
Added slider in layout editor

Jeff-Y-work:
https://github.com/user-attachments/assets/8cb16a61-ec2a-456d-9283-a0f196b0aa52

Current best try


han4wluc (approved):
you should take proper slider assets from route-graphics to test this.
i think there is small issues, but ok for now.

PR #546, opened 2026-02-13
feat/split slider direction and video loop

Jeff-Y-work:
- **Add option to choose if video should loop or not when adding background**
- **Split to horizontal and vertical slider; changed naming from idle to default**


src/components/commandLineBackground/commandLineBackground.handlers.js
```diff
@@ -149,6 +162,20 @@ export const handleFormInputChange = (deps, payload) => {
       tweenId: fieldValue,
     });
     render();
+    return;
+  }
+
+  if (name === "loop") {
+    const loopValue =
+      fieldValue === "true"
+        ? true
```
han4wluc: can we store directly as boolean?

han4wluc (approved):

src/components/commandLineBackground/commandLineBackground.handlers.js
Jeff-Y-work: Is it saying like store fieldValue as boolean `true`?

PR #591, opened 2026-03-02
feat/line editor paste

Jeff-Y-work:
- **feat(editor): add multi-line paste support in line editor**
- **Fixed lint**

https://github.com/user-attachments/assets/835ac78b-b01c-4e87-9be5-9eb1d458730b


han4wluc (approved):
looks good. just need to do testing for more edge cases.

like pasting in between lines, when there is content on top, bottom, with other actions, etc...

paste some weird content, wich maybe emoji, or characters from other langauge. or other weird characres.

Jeff-Y-work:
Tested with emoji, different language, different position...

Only found problem is \r\n for Windows, when pasted, \r remains as an invisible but existing character. Now this is fixed.

If no other problem, will merge soon.

PR #827, opened 2026-05-28
Add choice ordering support via context menu

maxchwa:
Added the option to toggle choice order.


han4wluc:
Screenshot 2026-05-30 at 22 43 37

if you right click on the 1st item. move up should not show.
if you right lick on the last item. move donw should not show.

han4wluc:
Screenshot 2026-05-30 at 22 48 26
Screenshot 2026-05-30 at 22 48 32

PR #828, opened 2026-05-30
Scene map changes

maxchwa:
- Added folder descriptions throughout RouteVN
(https://drive.google.com/file/d/1BMvyPD8BbIx_qcPVKzv9dGHYYd9igWAr/view?usp=sharing)

- Added the ability to pan to selected scene rectangles if they are not currently within the display (https://drive.google.com/file/d/1VA9FR2t3RfY3sxWreBPJeOkLzusdIw87/view?usp=sharing)


han4wluc:
next time better to split this into 2 PRs since they are not related to each other.

han4wluc:
for 'Added the ability to pan to selected scene rectangles if they are not currently within the display'
there was issue that if we jump directly to the selected rectangle box, it is too sudden. user may get lost in the map.
I've instead added a quick scroll, so it is more intuitive for the user on where on the map it moved.

PR #961, opened 2026-07-16
fix(project): patch legacy default menu styles

han4wluc:
## Summary

- patch legacy selected-menu font weight and Load text style during project open
- guard both partial updates by exact legacy values and record completion in the project KV store
- update the default repository template, persisted-key documentation, and regression coverage

## Validation

- bun run lint
- bun run test
- bunx vitest run tests/projectService/projectServiceCore.releaseRuntime.test.js
- bun test tests/project/layout.richText.test.js

### DIFF
--- a/docs/engineering.md
+++ b/docs/engineering.md
@@ -823,6 +823,17 @@ For the current full contract, see
 row value contract is in
 `docs/platform/12-windows-player-runtime-key-value-contract.md`.
 
+### Exceptional Project Content Patches
+
+Project content patches are narrowly scoped repairs for specific data written
+by a released RouteVN Creator template. They are not a general schema migration
+layer and must not bypass creator-model commands.
+
+The current marker, exact legacy-value predicates, no-op rules, remote-sync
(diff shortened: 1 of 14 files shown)


han4wluc:
Addressed the stale collaboration-state race in 876ce1e8. Web repository ensure now auto-applies content patches only for local projects. Remote projects trigger the patch only after initial sync completes, queued remote commands are projected, and no projection gap remains. Failed initial sync leaves the marker unset and performs no patch. Added regression coverage for successful sync, failed sync, local projects, and platform-level deferral.

han4wluc:
Added a dedicated platform contract for this exceptional repair in docs/platform/14-project-content-patches.md. It documents the exact IDs and legacy predicates, boolean marker lifecycle, no-op and retry behavior, partial command requirements, project-local marker scope, remote-sync safety gate, simultaneous-edit limitation, implementation/test mapping, and mandatory checklist for any future content patch. Linked it from the platform index, engineering guide, persisted-key catalog, and the non-obvious code boundaries.

PR #1116, opened 2026-08-27
feat: add scene reading time

JanLancelot:


han4wluc:
not needed to show reading time. line count and word count is enough.
reading time is too dependent on the user.
users can atually have tables on their own converting word count to reading time.

